PR #35, opened 2016-10-02
Proxy links for the python docs and read-only team repos

PeterJCLaw:


jimmythompson:
Gonna try the "rebase and merge", lemme know if this makes it a pain to keep up-to-date.

PeterJCLaw:
rebasing causes git's internal tracking not to know that this branch was merged, though the presence of the `Change-Id` lines in these commits (because I'm configured to add them to all my sr stuff) means that my script can track this.

PeterJCLaw:
Especially for cases where there is no merge to be had (e.g: it's a fast-forward) then I think we shouldn't rebase. I'd be happy to either merge or fast-forwards in such cases though.

PeterJCLaw:
Hrm, the python docs part of this doesn't seem to be working -- how can we:
- test it locally to debug?
- know that it's actually gone live?

jimmythompson:
As far as merging PRs goes, I'll stick to merge commits from now on.

---

To test the reverse proxy, you first need to generate an SSL private key (named `key.pem`) and a certificate (named `cert.pem`), and place them in a folder, so they can be mounted into the container.

If you then run these commands, you should have a working copy of the website on port 5679. 

``` shell
rake build
docker run -p 5678:443 -p 5679:443 -v (your secret folder):/etc/secrets --name srobo srobo/website
```

Making this process easier would be much appreciated.

---

Right now, the build process on commit generates a container and dumps it on [Docker Hub](https://hub.docker.com/r/srobo/website/). Someone (mainly me) then has to run a command to tell kubernetes to update itself.

The plan is to tag the images made in CI with the hash of the commit, then implement auto-updates, which would fix the manual deployment, and the "what's in live?" problem.

Shortly after your comment I updated the cluster, so it should be updated now.

PeterJCLaw:
Still 404s for me:
- https://studentrobotics.org/docs/python (should show the python 2.7 docs)
- https://studentrobotics.org/robogit-ro/SRZ (should challenge for your SR login)

PR #36, opened 2016-10-02
Remove timezones from events

PeterJCLaw:
Recording these dates with a timezone isn't entirely useful since
a reader then needs to work out whether BST applies on that date.
Remove the timezone information so we can just write dates in a
more natural form (also drop the redundant 'T' in the middle which
makes reading them harder).


jimmythompson:
Cheers for this, looks much better.

PR #37, opened 2016-10-03
Enable future mode when running the dev version of the site

Adimote:
Currently you can't see posts set to publish in the future when running a dev version of the site, this is a proposed fix for it, where you can see all future posts only for the dev site.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,5 +8,5 @@ task :deploy do
 end
 
 task :dev do
-  sh('bundle exec jekyll serve --drafts')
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
Hi,

Is there a way to have a custom (local, per developer) config as well?
Ideally in a way which doesn't error if it's not present?

I would advocate for the default to be a close to the real site as
possible, but then having instructions on how to add custom config, with
hints to the config keys developers might want to change.

Thanks,
Peter

Issue #34, opened 2016-10-02
Links from events to more information


PeterJCLaw:
At the moment the [events page](https://studentrobotics.org/events/) just shows the bare minimum of information about an event. While that is a useful reference, anyone actually coming to an event is going to want considerably more information.

In the past we've had a page dedicated to a given event, and the were linked from the old "key dates" page. I think we should aim for a similar system in the new website, albeit with a bit more structure to how we store the data.

With that in mind, I think we want a setup where:
- events are listed centrally, dynamically
- the information (both short- and long-form) for an event is all stored in a single place
- where possible the information about an event is recorded exactly once

Currently despite the minimal information available we already have duplication of the date of the event (in the file name and within the front matter). This is liable to cause copy-paste errors among other things, so we should sort it sooner rather than later.

PeterJCLaw:
cc @jimmythompson for design thoughts and @Adimote as this may affect the Kickstart pages

jimmythompson:
I had a quick stab at this for kickstart:

![screen shot 2016-10-07 at 08 33 20](https://cloud.githubusercontent.com/assets/653864/19182039/bc34c65c-8c68-11e6-8e2c-7133abbb3c98.png)

https://studentrobotics.org/events/sr2017/southampton-kickstart/
https://studentrobotics.org/events/sr2017/bristol-kickstart/
https://studentrobotics.org/events/sr2017/oxford-kickstart/

Each event has a content section which can contain any information we can't sensibly extract into metadata.

I think there's plenty of things we can flesh out in the metadata, (e.g. links to volunteer forms, and coordinates, etc.) which can power more calls to action and more dynamic elements of the page.

richardbarlow:
Hey Jimmy,

Those event pages look awesome. I see that you currently have a mailto link to teams@ when referring to the local team coordinator. That email address goes to the team coordinator, not any specific local team coordinator. We generally don't want teams contacting the team coordinator directly (other than in exceptional circumstances) as that puts an unmanageable load on a single person. The idea is that all communication with a given team is handled by their local team coordinator, who can pass information on if necessary. Obviously this makes things slightly more complicated for teams though. I think in the future it would be good to have a page hosted somewhere (it doesn't belong in the pr website) that lists teams and who their local team coordinator is. That page can then be linked to in situations like this.

In the mean time you could maybe remove the link around the existing text and add a footnote or some parenthesised text saying "[mailto]contact the team coordinator[mailto] if you don't know who this is"

[...]

PR #42, opened 2016-10-07
Fix events list render bug

PeterJCLaw:
Per discussion on #34 (though unrelated to that issue itself)


jimmythompson:
This appears to have shifted the navigation items down in the header:

screen shot 2016-10-08 at 20 24 54

PeterJCLaw:
That should be resolved in e7a6555.

@jimmythompson do you have any guidelines on the sorts of things which should be in each of the .scss files? Perhaps you could add some comments at the top of each and/or some central documentation to clarify that?

PeterJCLaw:
@jimmythompson This is becoming much more of an issue now that we're getting more events pages. Can you please take another look? I'm pretty sure the issue you raised with the SCSS affecting the header is fixed by https://github.com/srobo/website/pull/42/commits/e7a65552470f48c4df1eb6d866a47b8a1244d320.

_sass/base.scss
```diff
@@ -39,6 +39,14 @@ a {
   @include media-query("medium") {
     padding: $spacing 0 (3 * $spacing);
   }
+
+  ul {
```
jimmythompson: Moving this has messed some things around when I run this branch locally. Particularly on the home page.

PeterJCLaw:
Hi @jimmythompson good spot, thanks. I've given up on CSS change for now, so this branch just contains the major spacing fix.

Do you have some guidelines on which CSS classes should be used where? For example, the events pages are using `article-preview` for their main content blocks, which doesn't really make sense. Is there a general class we can apply to containers of wrapped prose text? Ideally that would be the parent of the block of SCSS I've been moving around here I think.

PeterJCLaw:
Meta: if you could ensure that in future you clarify exactly _what_ has changed (either in description or screenshots) it will make it much easier to correct things!

PeterJCLaw:
@jimmythompson ping on this.

jimmythompson:
Before I merge, can you just check you still get the problem in master? I played around with article styling, which included lists on event pages.

PeterJCLaw:
That does now seem to be fixed in the styling, thanks. (I still think that we should avoid having empty lines between list entries though).

PeterJCLaw:
No longer relevant.

PR #43, opened 2016-10-08
Various spelling fixes

PeterJCLaw:
Also includes a spelling config for VSCode.

### DIFF
--- a/.vscode/spellchecker.json
+++ b/.vscode/spellchecker.json
@@ -15,7 +15,6 @@
         "blueshirts",
         "Kickstart",
         "Kickstarts",
-        "KickStart",
         "Facebook",
         "YouTube",
         "online",
--- a/.vscode/spellchecker.json
+++ b/.vscode/spellchecker.json
@@ -1,5 +1,4 @@
 {
-    "language": "en_US",
     "ignoreWordsList": [
         "Brockenhurst",
         "Basingstoke",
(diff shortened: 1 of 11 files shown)


.vscode/spellchecker.json
```diff
@@ -0,0 +1,53 @@
+{
+    "language": "en_US",
```
RealOrangeOne: Should this not be en_GB?

.vscode/spellchecker.json
```diff
@@ -0,0 +1,45 @@
...
+        "Indersdorf",
+        "Mirfield",
+        "Tunbridge",
+        "prepend",
+        "baseurl",
+        "blueshirt",
+        "blueshirts",
+        "Kickstart",
+        "Kickstarts",
+        "KickStart",
```
RealOrangeOne: Is both lower case and upper case _'S'_ valid for _Kickstart_? Personally i'd only use lower

PeterJCLaw: We used to use "KickStart" quite heavily. I thought we'd removed all the usages of it, but there still seem to be some. Arguably we should not allow it since the chances of people editing the old files is quite low.

PeterJCLaw: 79d838f4592dd9bf18c99eb7ea971c3225ef9eb5

PR #46, opened 2016-10-12
Move the simulator zips to docs.git

richardbarlow:
This helps to isolate the docs from the PR website.


richardbarlow:
I have already added the simulator to docs.git and updated the links so it's safe to merge this right now.

Issue #29, opened 2016-10-01
Restore game archive


PeterJCLaw:
The old website had a page which detailed the games which have been played in the past. This served as a useful place for newcomers (volunteers, spectators and teams) to see what sorts of game are played, and so to get a good understanding of the competition itself.

For volunteers and teams it gives an idea of the scale of the challenge, so that they gain an understanding of how they can help and how hard it is to compete respectively.

For other visitors it provides a sense of scale (that we've been going as long as we have) as well as an idea of the challenge we present to the teams.

PeterJCLaw:
Possibly related to #3

PeterJCLaw:
Given the current structure of the website, this is actually going to live in the docs.
I'm currently in the process of [extracting the relevant files](https://gist.github.com/PeterJCLaw/c7527871f671b00854c036aae7939d2a) from https://github.com/srobo/srweb. I'll put up a PR shortly.

Issue #3, opened 2016-05-28
Competition pages


jimmythompson:
Something I'm keen on doing for this new design is having some sort of competition "archive". A set of pages for people to navigate back through each competition.

We could do a tournament tree, a league table, display some videos/photos and describe that year's game.

These pages would likely succeed the content on this page: https://www.studentrobotics.org/schools/game

![](https://i.stack.imgur.com/pnavf.jpg)

PR #48, opened 2016-10-12
Reducing the number of broken links in posts

edlongman:
Having looked at the broken links I have made a decision as to what to do with them.

Some have been removed if that is the best course of action

Some have been redirected to Wikipedia for game specific info for that year.

### DIFF
--- a/_posts/2014-10-26-student-robotics-2015-starts-with-a-range-of-new-kit.md
+++ b/_posts/2014-10-26-student-robotics-2015-starts-with-a-range-of-new-kit.md
@@ -21,7 +21,7 @@ The game: Capture the Flag
 This year's competition is all about retrieving flags from around the arena and putting them into the competitor's zone
 to get points. The flags are built from wooden cubes on ball casters.
 
-Read more about the [new game](https://en.wikipedia.org/wiki/Student_Robotics#2015:_.22Capture_the_Flag.22) or for full details please reference the rulebook (now archived)
+Read more about the [new game](https://en.wikipedia.org/wiki/Student_Robotics#2015:_.22Capture_the_Flag.22) or for full details please reference the rulebook (now removed)
 
 Kit upgrades
 ------------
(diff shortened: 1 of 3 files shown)


_posts/2014-10-26-student-robotics-2015-starts-with-a-range-of-new-kit.md
```diff
@@ -18,10 +18,10 @@ The game: Capture the Flag
 
 <img class="right" src="{{ site.baseurl }}/images/content/arena-2015.png" title="The 2015 arena layout" alt="An image of the 2015 arena layout">
 
-This year's competition is all about retrieving flags from around the arena and putting them into the competitor's zone 
+This year's competition is all about retrieving flags from around the arena and putting them into the competitor's zone
 to get points. The flags are built from wooden cubes on ball casters.
 
-Read more about the [new game](/schools/game) or for full details please reference the [rulebook](/resources/2015/rulebook.pdf).
+Read more about the [new game](https://en.wikipedia.org/wiki/Student_Robotics#2015:_.22Capture_the_Flag.22) or for full details please reference the rulebook (now removed)
```
richardbarlow: The rulebooks from previous years are not currently hosted anywhere. However, I think that we should probably add them back in either in this repo, or more likely the docs repo. They provide an interesting insight into the history of the games.

edlongman:
Maybe an archive section? Just where resources could be stored.

PeterJCLaw:
There's an outstanding ticket (#29; note the difference from #3 which
appears to be more about historic competition results) to add back an
archive page about the old games.

We should _not_ be linking to our own wikipedia page for this sort of
information (doing so looks terrible [worse than a dead link IMO] and
will likely cause wikipedia to remove the content).

edlongman:
Would it be useful to put the rulebooks for the years where there is missing info into the resources, #29  Possibly then the game pages as detailed in #3 could be built upon that.

PeterJCLaw:
Thanks for this, however many of these have since been fixed. Some remain, however it seems most likely that a fresh approach is going to be more efficient than trying to merge this PR given it's age.

PR #49, opened 2016-10-14
Add SR2017 Microgames

Adimote:
It's an urgent 'good to have'


richardbarlow:
For future reference, this sort of thing does not belong with the PR website. A better place for it would be the docs. An even better place would be as an artefact of a CI build that can be linked to.

Issue #50, opened 2016-10-22
Full stack local development instructions


PeterJCLaw:
There are currently no details on how to actually test the nginx config, meaning that changes to it are being made blind.

On #35 Jimmy mentioned the use of docker to do this (copied below), but it is non-obvious how complete this is, nor what other things are needed to have a complete local copy of the SR web presence.

The purpose of this ticket is to capture this need and gather the information needed to get started, with the aim of eventually adding some docs to the repo.

Jimmy wrote:

> To test the reverse proxy, you first need to generate an SSL private key (named `key.pem`) and a certificate (named `cert.pem`), and place them in a folder, so they can be mounted into the container.
> 
> If you then run these commands, you should have a working copy of the website on port 5679. 
> 
> ``` shell
> rake build
> docker run -p 5678:443 -p 5679:443 -v (your secret folder):/etc/secrets --name srobo srobo/website
> ```
> 
> Making this process easier would be much appreciated.

Questions to answer:
- what is `docker` and where do I get it from (is there any setup needed for it beyond installing it?)
- it looks like it's setting up two ports which both forward to the same target -- is that intentional?
- how does one generate the requisite SSL keys? (Can/should we use the keys from the [dummy-secrets](https://git.studentrobotics.org/cgit/server/dummy-secrets.git) repo)
- what is secret about the secret folder (and what goes in it? Is it just the SSL keys?)

jimmythompson:
> how does one generate the requisite SSL keys? (Can/should we use the keys from the dummy-secrets repo)

It'd be nice if this was automated as part of `rake build`, any self-signed cert will do. We could define the location of the secrets folder and ignore it in git.

> it looks like it's setting up two ports which both forward to the same target -- is that intentional?

No, my bad. 😅

One is forwarding port 80, one for 443.

> what is secret about the secret folder (and what goes in it? Is it just the SSL keys?)

For now, yes.

PeterJCLaw:
Ok, so I'm going to assume that we're talking about [this Docker](https://www.docker.com/) (which Ubuntu's repo calls `docker.io`) here. I've installed that.
I've then also done the hack where you need to add your user to a group `sudo usermod -aG docker $USER` so you can actually use the thing, complete with manadatory reboot (seriously, WTF it's 2016 -- reboots on install haven't been a thing for decades).

[...]

PR #53, opened 2016-10-26
Handle urls with slashes in a manner Gerrit copes with

PeterJCLaw:
This applies the same fix here as was applied in puppet at 370157e
(https://srobo.org/cgit/server/puppet.git/commit/?id=370157e84f6).

This should fix https://www.studentrobotics.org/trac/ticket/3023 _yet again_.

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -67,7 +67,7 @@ http {
     }
 
     location /gerrit/ {
-      proxy_pass https://saffron.studentrobotics.org/gerrit/;
+      proxy_pass https://saffron.studentrobotics.org$request_uri;
     }
 
     location /docs/python/ {


PeterJCLaw:
Note that I'm unable to test this since there is still no mechanism to do so.

richardbarlow:
Seems reasonable. Although I too have not tested this.

rspanton:
It is possible to run locally in docker to test.

PeterJCLaw:
> It is possible to run locally in docker to test.

I'm sure it is, but having spent an afternoon trying to do so (notes attached to #50) I couldn't get it working. @rspanton if you have some insight into how to get it working please could you comment on #50?

PeterJCLaw:
Thanks to @jimmythompson's new patches which make it possible to run docker more easily I've now verified that this does indeed fix the issue.

PR #54, opened 2016-11-05
Only regenerate the certificate if missing

PeterJCLaw:
This avoids the need to re-accept a new invalid certificate every time the container is run.


jimmythompson:
Looks good, cheers. 😄

PR #55, opened 2016-11-24
WIP - Announcing the first Southampton Tech Day

jimmythompson:
Not ready for merge yet, but starting the work for the Soton Tech Day.

@Adimote We need to confirm the location, also some other things which would be nice:

* a schedule for the day,
* directions to the venue (once confirmed, obviously!)
* is there a sign-up process?
* can any team attend?
* something for volunteers, unless you want to approach this differently. Might be worth asking Rob about Mailchimp.


Adimote:
This should close in favour of #56

jimmythompson:
Agreed, closing.

PR #61, opened 2017-02-08
Combine font css

PeterJCLaw:
This reduces the number of network round-trips and thus improves page-load times.


jimmythompson:
Didn't know you could do this, cheers Peter. 👍

PR #69, opened 2017-02-27
Add /comp redirect to SR2017 comp event page

richardbarlow:
This URL gets printed on the tickets. It's less typing (assuming anyone
actually bothers to type it in at all).

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -126,6 +126,9 @@ http {
     rewrite ^/feed.php             /feed.xml redirect;
     rewrite ^/password             /userman/ permanent;
 
+    # Ephemeral redirects (will probably change over time and/or not be useful after a point in time)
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
This looks fine, though I'd like to have a more concrete understanding of how the other competition pages are going to to fit into the url scheme before we merge this.

richardbarlow:
I have decided to put a shortened URL on the ticket instead. It's less typing for someone actually using it and it avoids any conflicts with the competition pages.

PeterJCLaw:
for clarity, the competition pages have previously included this redirection themselves, so I think that having this here is the right thing, I just wanted to understand what the plan is for the urls before committing us to anything.

(surely `srobo.org/comp` is pretty short anyway?)

PR #71, opened 2017-03-18
Competition timings

PeterJCLaw:
This brings the competition photo and knockouts forwards. This allows for a little bit more spare space in the knockouts, which have historically overrun.


PeterJCLaw:
cc @richardbarlow

PeterJCLaw:
Note that the league matches this year are expected to finish at 14:15 (with any delays potentially pushing that later).

richardbarlow:
I definitely appreciate this shuffling of photo/knockouts start as it gives me 10 minutes in which to perform the arena changeover, rather than the ~2 minutes that was scheduled previously.

richardbarlow:
Can this change be merged in?

PR #72, opened 2017-03-26
Add misisng steps for working on the reverse proxy

PeterJCLaw:
More information about how to get going using docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,12 @@ the reverse proxy is working before you commit. To do this, do the following:
 
 1. [Install Docker][docker]
 
-2. Start the container
+2. Fetch the latest base image:
+    ``` shell
+    $ docker pull nginx
+    ```
+
+3. Start the container
 
     ``` shell
     $ rake run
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,12 +28,35 @@ the reverse proxy is working before you commit. To do this, do the following:
 
 1. [Install Docker][docker]
 
-2. Start the container
+2. Fetch the latest base image:
+    ``` shell
+    $ docker pull nginx
+    ```
+
+3. Optionally change the bound ports so they don't conflict with your local host's
```
jimmythompson: Maybe a different Rake task would make this easier?

PeterJCLaw: Does the `Rakefile` get used in production? If not, I'm tempted to just change the run task.

PeterJCLaw: Ok, I'll have a think about how best to do that. I think it's probably best to approach that separately though, as there is already enough info here for people to get going and port 80 already being in use may just be an edge case.

PR #73, opened 2017-03-26
Proxying for the competition website 

PeterJCLaw:
This includes the /comp pages which can go live immediately (while they're not quite ready, they're not yet linked from anywhere and there's no reason to keep them hidden before the competition).

This also includes the override for the competition homepage, which is deliberately commented out so that the current homepage will remain for this week.

Once this PR is merged I'll open another one that un-comments the competition homepage override which can be merged on the morning of the competition.

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -108,6 +108,18 @@ http {
       proxy_pass https://saffron.studentrobotics.org/comp-api/;
     }
 
+    # Provide access to the competition pages under the normal prefix
+    location /comp/ {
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
Is this live yet? I was hoping to validate it, but visiting https://studentrobotics.org/comp/ returns a 404 rather than the page I was expecting.

jimmythompson:
Just pushed, should be working now.

PeterJCLaw:
Yup, seems to be, thanks.

PR #74, opened 2017-03-29
Ensure that /comp/ links are cleaned

PeterJCLaw:
This has us rewrite the links to the individual pages so that they all end up at `/comp/$page` rather than `/competition-website/comp/$page` as previously.

Fixes srobo/competition-website#3.


PeterJCLaw:
See also 909ef38 which applies an equivalent change to the homepage block.

PeterJCLaw:
@jimmythompson thanks! please can you update this once it' live so I can validate the fix.

PeterJCLaw:
confirmed fixed.

PR #75, opened 2017-03-29
Make the competition homepage live

PeterJCLaw:
**Do not merge until Friday evening at the earliest**

PR for making the competition homepage live.

This also includes some fixes to make the relevant (otherwise commented out) block actually work. Ideally when we revert this we will only revert the commit which removes the comments.

NB: if there's a better way to make this easier to enable/disable I'd be happy to hear it. I don't like commented code.

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -110,18 +110,18 @@ http {
 
     # During the competition we un-comment this block to override the homepage
     # with the comeptition-specific one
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
cc @jimmythompson @richardbarlow.

PeterJCLaw:
cc @thomasleese

PR #83, opened 2018-09-22
Spelling en gb

PeterJCLaw:
Depends on #82.


PeterJCLaw:
Ideally we'd have a spell-check solution which is present in any editor you like.

Issue #79, opened 2018-09-19
Refresh the brand


PeterJCLaw:
There's a new SR icon and colour, we need to update the pages to pick them up. We should do this at some point, but it's not blocking anything.

@jimmythompson a couple of questions:
- where's the original media for the new icon?
- does the website get the icon directly from the style repo, or does it have its own copy? (I can probably work this one out, but if you happen to know the answer it'd be useful to know)

PeterJCLaw:
@Adimote I think you've ended up doing this in #81? If so, could you co-ordinate with @RealOrangeOne (as he's assigned to this).

PeterJCLaw:
Ok, from what I can tell, #81 covers the icon update but not the colour update. The colour update is dependent on https://github.com/srobo/brand/pull/3.

Adimote:
#85 fixes this, it was previously #81 but I split the PR

PeterJCLaw:
Done as the related PRs have all merged, including https://github.com/srobo/docs/pull/55 which updates the docs.
I pushed changes to update the IDE separately; while it doesn't have the right fonts, it now has the right colours (which is more than it's ever had in the past!)

PR #85, opened 2018-09-22
Update to the new logo

Adimote:
Change the logo to the new one.

Partially fixe.s #79


PeterJCLaw:
Looks fine, out of interest where did the `png`s come from?

Adimote:
I generated them with inkscape, plus some manual munging to find the right size to fit the logo cropping that happens when you're in mobile view

jimmythompson:
Cheers Andy, it LGTM, but I've just noticed that the logo is quite blurry on a high resolution display.

Please could you try converting the logo in something like imagemagick and documenting that process, so if I wanted to produce a @2x of the logo, it's fairly predictable? I can then roll this process back into srobo/style.

If its a pain wrangling with imagemagick, it's no problem, we can merge anyway and raise an issue.

Thanks!

Adimote:
My experience with Imagemagick is it can often render SVGs really badly. Inkscape seems to be the standard way of automating it, I'll write a script/docs to do it.

Adimote:
I've made a script and a description, but I can't think of a nice place to put it, should I put it straight into the style repo?

PeterJCLaw:
@Adimote yeah, I'd suggest putting it in https://github.com/srobo/style. I almost put a `Makefile` in there myself when updating the IDE this afternoon, so a script which you can pass in a size to would be handy. Alternatively, just having an example which is easy to copy/vary would be useful.

Here's what I ended up using: `inkscape assets/logo-white.svg --export-png=$HOME/public_html/sr/ide2/web/images/static.png --export-width=52 --export-height=52`.

I think we probably want to get into the habit of outputting images at 1x, 2x and 3x resolutions and then using [`srcset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#attr-srcset) or [similar mechanisms](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) to scale up the images.
The convention I've seen used for this sort of thing is to have multiple copies of the image in the repo, named something like: `logo.png`, `logo@2x.png`, `logo@3x.png`.

PR #87, opened 2018-09-22
Update the sign-up forms to be the new mailchimp lists

### LINKED ISSUE #88: Update the sign-up forms to be the new mailchimp mailing lists
The current forms are currently unofficially for SR2017, we've created new mailing lists for SR2018, so we should use them.

Adimote:
fixes #88


_includes/team-signup.html
```diff
@@ -7,31 +9,72 @@
...
-      <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="b_1af17c2a3426de258620aae46_dd7d04139d" tabindex="-1" value=""></div>
-      <div class="clear">
-        <input class="button button-primary" type="submit" value="Register your interest" name="subscribe" id="mc-embedded-subscribe">
-        <span>or</span>
-        <a class="button" href="mailto:{{ site.emails.teams }}" target="_blank">Speak to someone</a>
+     <label for="mce-LNAME">Last Name </label>
+     <input type="text" value="" name="LNAME" class="" id="mce-LNAME">
       </div>
+<div class="mc-field-group size1of2">
+        <label for="mce-PHONE">Phone Number </label>
```
PeterJCLaw: Do we really need to ask for phone number and postal address here?

As I understand it, this form is for people who are vaguely interested and would like to know more, rather than a full commitment to leading a team soon.

Given that, this level of detail feels like to put people off.

_includes/volunteer-signup.html
```diff
@@ -1,23 +1,50 @@
...
+	<label for="mce-LNAME">Last Name </label>
+	<input type="text" value="" name="LNAME" class="required" id="mce-LNAME">
+</div>
+<div class="mc-field-group">
+	<label for="mce-MMERGE3">Location </label>
+	<input type="text" value="" name="MMERGE3" class="" id="mce-MMERGE3">
+</div>
+<div id="mergeRow-gdpr" class="mergeRow gdpr-mergeRow content__gdprBlock mc-field-group">
+    <div class="content__gdpr">
+        <label>Consent</label>
```
PeterJCLaw: Do we really need an explicit consent box?

They're giving us their information explicitly so that we can contact them about volunteering. We're not going to _market_ things to them (if we were then they would need to opt in to that), we're only going to tell them about things they're directly asking to be told about.

css/main.scss
```diff
@@ -20,6 +20,5 @@ $baseurl: "{{ site.baseurl }}";
         "sponsors",
         "footer",
         "tags",
-        "splash",
```
PeterJCLaw: This change seems unrelated.

volunteer.html
```diff
@@ -70,8 +70,8 @@ <h3 class="text-center">Development</h3>
 <div class="panel">
   <div class="panel-container">
     <div class="panel-fill text-center row">
-      <div class="column l-6-12 l-offset-3-12">
         <a name="signup"></a><h2>Want to volunteer?</h2>
+      <div class="column m-8-12 m-offset-2-12 text-left">
```
PeterJCLaw: Please can we fix the indentation of the previous line.

_includes/volunteer-signup.html
Adimote: This form was generated using the mailchimp form generator, the consent checkbox can't be turned off, so I suspect their lawyers have looked into this and made this decision

css/main.scss
Adimote: oops, left over from a previous change

_includes/team-signup.html
Adimote: Currently all of these fields are set as 'required' in mailchimp, so I can't change this here.

@jimmythompson is there any reason why the phone number and address aren't optional?

Adimote: Oops just looked into this, the phone number is actually optional, the point still stands for address however

_includes/volunteer-signup.html
PeterJCLaw: Most likely MailChimp will be assuming that we're collecting the data to use it for marketing purposes. We're not. As I understand it, GDPR consent only applies to marketing messages.

volunteer.html
Adimote: fixed in 5a1219a220d4e74aa963c6b5d7bd4c34643cc123

_includes/volunteer-signup.html
Adimote: I'm not certain I agree with you, but I'll remove it anyways because I trust your judgement and really don't want to spend another evening reading GDPR.

PeterJCLaw: Is there not an option in the MailChimp generator to say that we're not collecting _marketing_ emails?

PeterJCLaw: I think @prophile may have more GDPR understanding than I and thus may have an opinion here.

_includes/team-signup.html
```diff
@@ -1,4 +1,6 @@
-<form action="//studentrobotics.us12.list-manage.com/subscribe/post?u=1af17c2a3426de258620aae46&amp;id=dd7d04139d"
+<!-- Begin MailChimp Signup Form -->
+<div id="mc_embed_signup">
+<form action="https://studentrobotics.us12.list-manage.com/subscribe/post?u=1af17c2a3426de258620aae46&amp;id=3ce424e7fa"
```
RealOrangeOne: I believe the `&` is incorrectly escaped here?

Adimote:
Oops looks like I merged in a bunch of changes in the latest push, will fix

Adimote:
I've rebased it on top of master to fix conflicts, it should be good to go now

PeterJCLaw:
@RealOrangeOne that's what local development environments are for, though I agree live linked previews are much nicer.

PR #93, opened 2018-09-27
Update the banner button to direct to the compete page

Adimote:
The current home button leads to /about which is not as useful to possible competitors as /compete.

Change the button to redirect to /compete


PeterJCLaw:
While I appreciate the sentiment here, I'm not sure this is a good idea -- the button on the home page needs to make sense for all our audiences, which includes volunteers and sponsors too.

I'd be up for having _more_ buttons though, as long as it doesn't start feeling cluttered. We might want to think about how they relate to the links in the top right too if we added more.

Adimote:
Gonna throw this away as we don't want to direct people to Compete

PR #96, opened 2018-09-28
Update README for newcomers

PeterJCLaw:
This adds things found while setting up the site for a new developer (on Windows).

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ The Student Robotics public website.
     $ gem install bundler rake
     ```
 
-3. Start the app in development mode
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
Having fixed #99, I believe the addition of `npm` being required turns out to be the only thing that we needed to add.

BKiddo:
Sorry to be an idiot but I'm not sure that I have enough (any) web development fluency to reasonably review this?

BKiddo:
Sorry I am rather useless when it comes to anything web!

PR #97, opened 2018-09-28
Correct information text - entries are currently open

PeterJCLaw:


compete.html
```diff
@@ -130,10 +130,7 @@ <h3>Building a robot</h3>
       <a name="signup"></a><h2>Want to compete?</h2>
       <div class="column m-8-12 m-offset-2-12 text-left">
         <p>
-          Entries for this year's competition have now closed, but if you would
-          like to register your interest in applying for a place in a future
-          competition, please enter your details below and we'll contact you once
-          applications for next year's competition have opened.
+          Entries for the SR2019 year competition are open until the 26th October 2018.
```
jimmythompson: The phrase "for the SR2019 year competition" sounds odd. Maybe "for SR2019", or "for this year's competition"?

PeterJCLaw: good point

PeterJCLaw: 0306534 fixes this

I wanted to try to be explicit about the year, but got in a mess. This is the simplest solution and it's clear for now. We can improve this later.

PR #98, opened 2018-09-29
Rename config key for recent Jekyll change

PeterJCLaw:


RealOrangeOne:
What's the context behind this change? We've not changed the jekyll version in a while, so i'd assume this works either way round. Is using `gems` deprecated?

PeterJCLaw:
When I run `rake dev` I get the following deprecation warning:
```
bundle exec jekyll serve --drafts --config _config.yml
Configuration file: _config.yml
       Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.
```

I had assumed that we'd started seeing this as a result of a recent jekyll upgrade - I don't recall seeing this previously.

PR #103, opened 2018-10-01
Be explicit that places aren't guaranteed

RealOrangeOne:
It's unlikely we'll need to prevent teams from completing, but we should still account for it!

### DIFF
--- a/_posts/2018-09-27-sr2019-announced.md
+++ b/_posts/2018-09-27-sr2019-announced.md
@@ -32,9 +32,9 @@ Rising to the top of the knockout stages is not the only way to impress the
 judges though; prizes will be available in a number of other categories to be
 announced at kickstart.
 
-If you would like to compete in Student Robotics 2019, please fill in the
+If you would like a chance to compete in Student Robotics 2019, please fill in the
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
@PeterJCLaw I don't see this copy on the `complete` screen, nor anything like it which pertains to the fact places aren't 100% confirmed.

PeterJCLaw:
Ah, I was thinking of the line:
>  Entries for this year's competition are open until the 26th October 2018. 

though you're right that that is already framed differently than this was.

jimmythompson:
Can you elaborate a little bit on what happens if you'd like to enter multiple teams?

RealOrangeOne:
@jimmythompson currently it's just stored in MailChimp. When it comes to selection, I'd argue we try and fit all teams who requested a place in as we can. If we have the ability to allow more places, then we should allocate second teams on 'first come, first served'

PeterJCLaw:
@RealOrangeOne I think that @jimmythompson meant: could we change the wording to be a bit more explicit about what happens if one wants to enter multiple teams.

PR #113, opened 2018-11-17
Move to Circle CI 2.x config

### LINKED ISSUE #77: Port CircleCI config to v2
The deadline for switching over is getting rather close: https://circleci.com/sunset1-0/

PeterJCLaw:
Fixes #77.


PeterJCLaw:
Huh, that force push thing is new! (for the record, I'm trying to work out why Circle CI isn't building that last commit).

.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
+version: 2
+
+jobs:
+  build_only:
+    docker: &docker-template
```
RealOrangeOne: I have no idea what this syntax means. What's it doing?

PeterJCLaw: It's a YAML anchor -- it provides a name which refers to the content of the current key and which can be used to "copy" that content into another key later on.

README.md
```diff
@@ -10,8 +10,11 @@ The Student Robotics public website.
 
 2. Install Bundler and Rake
 
-    ``` shell
-    $ gem install bundler rake
+    ```
+    $ cat ./scripts/install-gems
```
RealOrangeOne: Why are we telling people to `cat` the file before running it?

PeterJCLaw: So they know what its going to do. I'm happy to do that if you think it's not useful.

PeterJCLaw: I'm pretty sure that most people won't _actually_ cat it.

RealOrangeOne: I think i'd rather we still not include this. If people want to know what's going on, they can just open the file themselves.

scripts/install-gems
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
RealOrangeOne: Can we use `/usr/bin/env bash` here, and ensure we do `set -e` below, so it fails with the right exit code

PeterJCLaw: I don't want this to be tied to `bash`, so I'd rather leave the shebang alone. I'll change the line below to `exec` though so that we get the right exit code behaviour.

PeterJCLaw: d02b345

README.md
PeterJCLaw: 39cdd25

PR #117, opened 2018-12-14
Add short-hand access to the rules

RealOrangeOne:
Add a short-hand URL for accessing the rules, rather than srobo.org/docs/rules, or srobo.org/docs, then clicking rules.

If my understanding of the setup is correct, this should also allow srobo.org/rules to redirect to it.

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -81,10 +81,6 @@ http {
     location /docs/python/ {
       proxy_pass https://patience.studentrobotics.org/docs/python/;
     }
-
-    location /rules/ {
-      proxy_pass https://patience.studentrobotics.org/docs/rules/;
-    }
     
     location /docs/ {
       proxy_pass       https://srobo.github.io/docs/;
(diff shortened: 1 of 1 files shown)


_env/nginx.conf
```diff
@@ -82,6 +82,10 @@ http {
       proxy_pass https://patience.studentrobotics.org/docs/python/;
     }
 
+    location /rules/ {
+      proxy_pass https://patience.studentrobotics.org/docs/rules/;
```
PeterJCLaw: I'm not sure we can spell this as we want using a `location` block. I think we want a `rewrite` tagged as a redirect (there's already a bunch of them lower down).

I think we want something like:
``` nginx
    rewrite ^/rules                /docs/rules permanent;
```

_env/nginx.conf
```diff
@@ -81,7 +81,7 @@ http {
     location /docs/python/ {
       proxy_pass https://patience.studentrobotics.org/docs/python/;
     }
-
+
```
PeterJCLaw: unrelated change?

RealOrangeOne: Viiiiiiimmmmmm!

PR #121, opened 2018-12-14
Cancel the London December Tech Day event

PeterJCLaw:
Remove the directions and schedule from the page as they no longer apply, but leave the introduction in place and add a bold statement that the event is not running.

This replaces #118.

*This event page*:
image

*All events page*:
image

### DIFF
--- a/_events/sr2019/london-tech-day-december.md
+++ b/_events/sr2019/london-tech-day-december.md
@@ -4,6 +4,7 @@ date: 2018-12-15 10:00:00
 layout: event
 type: techday
 location: Thread, Whitechapel
+cancelled: true
 ---
 
 Tech Days are opportunities for teams to get help if they're struggling with a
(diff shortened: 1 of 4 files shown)


_events/sr2019/london-tech-day-december.md
```diff
@@ -16,32 +16,8 @@ limited to six teams, so be sure to [let us know][teams-contact] if you'd like
 to attend. For this Tech Day, please express your interest by Wednesday 12th
 December.
 
-**Please bring your laptops!**
+~~Please bring your laptops!~~
```
RealOrangeOne: Why are we changing this to be crossed out?

PeterJCLaw: It doesn't make sense to leave it in place, but this felt a better solution than removing it. Happy to remove it if you think that's better.

RealOrangeOne: I'd argue leaving it as-is, it feels unnecessary to change it

PeterJCLaw: My concern was that leaving it as it was could be confusing as it somewhat implies the event is going ahead.

PeterJCLaw: Removed entirely in https://github.com/srobo/website/pull/121/commits/9e05007c630176ea45eddc28c96fe59a80d61ec2 as that seems simplest.

Issue #52, opened 2016-10-22
srobo.org/docs/kit redirects to srobo.github.io/docs/kit/


PeterJCLaw:
If you visit http://srobo.org/docs/kit then you end up at https://srobo.github.io/docs/kit/.
This appears to be related to the redirection which happens when there isn't a trailing slash in the original URL, as http://srobo.org/docs/kit/ works correctly. Note that this can be reproduced with arbitrary pages, for example http://srobo.org/docs/rules also shows this.

Clearly this is undesirable as people end up outside our web domain.

PeterJCLaw:
This was indeed fixed by #123.

Issue #122, opened 2019-01-19
Nav bar on docs section leads to 404


shardros:
When in the a section of the site where the URL is https://srobo.github.io/ the links in the navy bar point to the wrong places, giving a 404

shardros:
I went to https://studentrobotics.org/docs/ then touched `programming` and I was redirected to srobo.github.io

PeterJCLaw:
Aha, thanks.

This turns out to be an issue in how GitHub Pages handles redirects. I've reported the issue to GitHub, though also have a potential fix for our proxy config.

While that doesn't admittedly solve the issue if you do manage to reach `srobo.github.io`, it should make it much less likely that anyone actually does.

shardros:
Ah okay, out of interest why do you have the GitHub pages site as well as the main site?

PeterJCLaw:
We use GitHub Pages for hosting and proxy that into our own domain. Pages is great for hosting as it means that we (mostly) don't need to worry about the build/deploy steps ourselves.

Adimote:
https://github.com/srobo/docs/pull/80 was merged and shipped, which fixes the links, so I'm gonna declare this fixed

PeterJCLaw:
https://github.com/srobo/docs/pull/80 fixes some specific _causes_ of this, though neither fixes the actual issue, nor the general case cause. The actual links may not be worth fixing (since we don't really want people to get to the srobo.github.io pages anyway), though I think the general case of the cause is.

I've put up #123 which improves our handling of GitHub's urls via a tweak to our proxy config, applied across all the github.io domains we use.

This was the bug which I'd reported to GitHub. I'd originally been concerned that it might be considered a security issue (hence the obliqueness about it in my earlier comments), but they've confirmed that they don't consider it sensitive.

Essentially the issue is that when redirecting from from `/docs/programming` to `/docs/programming/` a redirect is issued to `http://srobo.github.io/docs/programming/` (note the `http://` protocol). As that (insecure) url is different to the url which our proxy is expecting, the proxy misses it.

PR #123, opened 2019-01-24
Ensure redirect consistency even when GitHub pages isn't

PeterJCLaw:
GitHub pages issues redirects to canonical urls in the form of insecure urls (which then upgrade when requested). I'm assuming that this is a bug and have reported it to GitHub. Specifically, they canonicalise from `/docs/programming` to `/docs/programming/` (which is a good thing overall) by issuing a redirect to `http://srobo.github.io/docs/programming/` (note the `http://` protocol). This is insecure :( and is different to the url which our proxy is expecting, meaning that the proxy misses it.

The fix here is to detect the insecure urls in the redirects and proxy those too.

Further mitigates https://github.com/srobo/website/issues/122

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -85,6 +85,13 @@ http {
     location /docs/ {
       proxy_pass       https://srobo.github.io/docs/;
       proxy_set_header Host srobo.github.io;
+      # Work around github.io issuing HTTP (not HTTPS) redirects when
+      # redirecting to cannonical forms of urls. For example, given the url
+      # "https://studentrobotics.org/docs/programming", which we proxy to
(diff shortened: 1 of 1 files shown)


antoinepetty:
I'm not sure how I can test this to ensure it works, but I can't see any obvious errors in the changed file.

PeterJCLaw:
@antoinepetty @RealOrangeOne The way to test this involves running the proxy locally and checking that the redirects are properly converted. The README has instructions for [working on the proxy](https://github.com/srobo/website/blob/master/README.md#to-work-on-the-reverse-proxy), though if there's improvements we can make which would improve it please do suggest things.

PeterJCLaw:
I suspect this will also fix #52.

PR #126, opened 2019-02-07
Change tech day map to point to B53

antoinepetty:
Swapped PDF of Southampton tech day map with one which points to L2 labs.


PeterJCLaw:
It's great that this has gone live, however in future I would be strongly in favour of creating a new map rather than editing an existing one. Doing so avoids creating any confusion as the previous tech day event still links to this map (also we now have to remember edit this back for the March Tech Day if that event doesn't also get moved).

antoinepetty:
> It's great that this has gone live, however in future I would be strongly in favour of creating a new map rather than editing an existing one. Doing so avoids creating any confusion as the previous tech day event still links to this map (also we now have to remember edit this back for the March Tech Day if that event doesn't also get moved).

Good point - will do this from now on.

PeterJCLaw:
I've done that in https://github.com/srobo/website/pull/127, though it's not urgent that that ships.

PR #127, opened 2019-02-08
Separate out the maps for Zepler than Building 25

PeterJCLaw:
This adds a small amount of local information to the Tech Days in known locations and restores the previous map while moving the new one to its own name.


PeterJCLaw:
> Can we rename the `southampton-techday-directions-highfield.pdf` file to something which explicitly states it's building 25, just so it's more obvious.

I wondered about that. Certainly it would avoid people accidentally downloading the wrong one if they'd followed the link.

PR #128, opened 2019-02-09
Copy navigation toggle from docs

as2247:
Changed toggle code to match docs toggle code. Fading out has been lost (can be added back if needed).
image

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -4,7 +4,7 @@
       <li><a id="navigation_toggle" href="#"><i class="fa fa-fw fa-bars"></i></a></li>
     </ul>
   </div>
-  <ul id="navigation" class="full-navigation">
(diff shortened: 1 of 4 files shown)


Adimote:
Does this mean we can get rid of jquery?

Adimote:
Looks like we can, jut remove 
```
<script src="https://code.jquery.com/jquery-3.0.0.min.js"
        integrity="sha256-JmvOoLtYsmqlsWxa7mDSLMwa6dZ9rrIdtrrVYRnDRH0="
        crossorigin="anonymous"></script>
``` 
from `includes_/scripts.js` and it's gone!

Thanks for doing this, Alana!

as2247:
> Does this mean we can get rid of jquery?

already on it

Adimote:
Removing jquery will speed up the site load time by 180ms :tada:

PR #130, opened 2019-02-10
Drop condition on depedency

PeterJCLaw:
This means that the gem will be installed on all platforms, avoiding the `Gemfile.lock` changing between platforms.


PeterJCLaw:
Have checked on windows -- this still works fine there (admittedly tested against an existing install) and also does not modify the Gemfile there too.

PR #137, opened 2019-03-16
Add a mailmap

PeterJCLaw:


PeterJCLaw:
I'm ideally hoping for confirmation from each of those affected (hence the inclusion of all as reviewers), though I might just get this merged when we've had a couple more reviews.

PR #142, opened 2019-03-27
Switch to srcomp.studentrobotics.org for the comp-api

PeterJCLaw:
This is the public-facing compbox.

Related to https://github.com/srobo/tasks/issues/233 and https://github.com/srobo/tasks/issues/112

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -111,7 +111,7 @@ http {
     }
 
     location /comp-api/ {
-      proxy_pass https://saffron.studentrobotics.org/comp-api/;
+      proxy_pass https://srcomp.studentrobotics.org/comp-api/;
     }
 
     # During the competition we un-comment this block to override the homepage


PeterJCLaw:
> This code change looks fine. Is there a reason this isn't hosted on `patience`? (Besides load bearing)

Primarily load bearing. `saffron` did run this in the past (and still does) and `patience` could easily run this, but having separate machines simplifies things.

There's also the benefit of the two compboxes being the same config -- no need to worry about needing to keep their configs in sync manually. If we did need to change anything on the day, it ought to be trivial to apply to the other one.

I'd also quite like to not have the main server puppet config do srcomp things explicitly. Ideally the srcomp puppet stuff would be a re-usable module, such that if we did want it on the same box we'd just include it like any other module. Unfortunately it's not really in a place where that would work.

What I haven't yet worked out is how we keep the pages usable after the competition. We can keep the second machine around for a while, but it's not an ideal longer term solution. Maybe at that point we'll do the work to make the srcomp puppet properly re-usable, or maybe we'll switch back to using the version on `saffron` (though the latter is _far_ from ideal).

PeterJCLaw:
Oh, there is one other, rather key, reason -- the config on `saffron` (which would also be what I'd deploy to `patiences` if we did that) doesn't include the stream or any of the in-venue screens. Adding those is possible, but nontrivial and since I'd like to move in the direction of `srcomp-puppet` being the primary isn't really worth it.

PR #149, opened 2019-04-16
Update this year in sign-up form

Adimote:


compete.html
```diff
@@ -130,7 +130,7 @@ <h3>Building a robot</h3>
       <a name="signup"></a><h2>Want to compete?</h2>
       <div class="column m-8-12 m-offset-2-12 text-left">
         <p>
-          Entries for this year's competition are open until the 26th October 2018.
+          Entries for Student Robotics 2020 are open until around September 2019.
```
PeterJCLaw: Should we perhaps call this "Applications" to help further clarify that not everyone gets a place?
Just a thought.

RealOrangeOne: We should definitely do this!

compete.html
```diff
@@ -130,7 +130,7 @@ <h3>Building a robot</h3>
       <a name="signup"></a><h2>Want to compete?</h2>
       <div class="column m-8-12 m-offset-2-12 text-left">
         <p>
-          Entries for this year's competition are open until the 26th October 2018.
+          Applications for Student Robotics 2020 are open until around September 2019.
```
RealOrangeOne: This form signs them up to the master list, so isn't actually opening them to SR2020. The way we've done it in the past is signed people up to the master list now, and then in September email out an actual signup link, and update the one on the website.

PeterJCLaw:
Closing as this is redundant now that we've opened applications for SR2020.

PR #150, opened 2019-04-29
Restore year-agnostic signup forms

RealOrangeOne:
Revert https://github.com/srobo/website/pull/87/ to revert back to year-agnostic signup forms. This affects both volunteer and team leader forms.

I've tested both work locally, and put details into the list i'd expect. 

I appreciate this is almost impossible to reasonably review.

Replaces https://github.com/srobo/website/pull/149

### DIFF
--- a/volunteer.html
+++ b/volunteer.html
@@ -70,8 +70,8 @@ permalink: /volunteer/
 <div class="panel">
   <div class="panel-container">
     <div class="panel-fill text-center row">
-      <a name="signup"></a><h2>Want to volunteer?</h2>
-      <div class="column m-8-12 m-offset-2-12 text-left">
+      <div class="column l-6-12 l-offset-3-12">
+        <a name="signup"></a><h2>Want to volunteer?</h2>
         <p>Sign up and we'll keep you updated on volunteering opportunities.</p>
         {% include volunteer-signup.html %}
       </div>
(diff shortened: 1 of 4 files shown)


volunteer.html
```diff
@@ -70,8 +70,8 @@ <h3 class="text-center">Development</h3>
 <div class="panel">
   <div class="panel-container">
     <div class="panel-fill text-center row">
-      <a name="signup"></a><h2>Want to volunteer?</h2>
-      <div class="column m-8-12 m-offset-2-12 text-left">
+      <div class="column l-6-12 l-offset-3-12">
+        <a name="signup"></a><h2>Want to volunteer?</h2>
```
PeterJCLaw: I don't think we want these changes?

RealOrangeOne: It doesn't seem to make a difference to the output, but the markup being like this is definitely more correct.

PeterJCLaw: While I can't say exactly what it fixes, it does seem that this change was deliberate (if possibly unrelated): https://github.com/srobo/website/pull/87/commits/81a9325f1f7cff7a83ebe09ac9800521fbc1fae1 and https://github.com/srobo/website/pull/87/commits/5a1219a220d4e74aa963c6b5d7bd4c34643cc123

RealOrangeOne: The main change is the addition of a `text-left` class, but the text it meant to be centered anyway

PR #154, opened 2019-07-23
Drop HTTPS as part of the move to DigitalOcean

jimmythompson:
Hey all,

As part of the certificate expiry problem I've started to move this off Google Cloud and onto DigitalOcean. Like our existing infrastructure, it's still in a container, and there's still a kubernetes cluster.

So the way it's going to work for the time being is that:

* https://studentrobotics.org points to a DigitalOcean load balancer, which balances across all the Kubernetes nodes. Right now there is only one node.
* On each of those nodes the website runs on port 30000.
* The load balancer terminates SSL, the certificate is renewed and managed by DigitalOcean & LetsEncrypt.

There's a working example of everything I've setup so far at https://studentrobotics.co.uk, with a forked Docker image running these changes. I've also got some Terraform code, which provisions the DigitalOcean account and the cluster, coming in another repository.

Next steps will be to pull the website onto some free static hosting, like we do with the docs, and have a separate repository for the reverse proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,11 @@ the reverse proxy is working before you commit. To do this, do the following:
     $ rake run
     ```
 
-5. Visit <https://localhost:443> in a browser (adjust the port as needed if you
+5. Visit <http://localhost> in a browser (adjust the port as needed if you
    edited it), and confirm that you see a copy of the SR website (also check that
    you didn't get redirect to the real one!)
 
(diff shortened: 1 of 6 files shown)


jimmythompson:
Also, Docker/Kubernetes isn't set in stone longer term, it's just what it's written in now and we need something in place before it expires!

PR #176, opened 2019-10-18
Add more specific details for Southampton campus

RealOrangeOne:
Add the location of the intro talk for the Southampton kickstart.

The PDF is a modification of the one used in 2019, with the locations changed.

Not _entirely_ sure on the address, but most people really only use the postcode!


trickeydan:
Please could you update the map to a current map. That map shows parking where there **is not** parking.

You can get a current copy on maps.soton.ac.uk

You can also get embedded maps from the same site.

RealOrangeOne:
@trickeydan could you be more specific? Looking at the map on `maps.soton.ac.uk`, I think there's parking in all 3 parking locations.

Specifically which ones aren't parkable? We definitely don't want to be telling teams to park in places there isn't parking!

TomWheal:
Oh yeh, building 100 isnt shown either. The area we have labeled as parking are all still valid, its just other ones that are no longer valid

PR #177, opened 2019-10-21
Fix any_events when a cancelled event is the only one

PeterJCLaw:
This ensures that if we have a cancelled event as the only future event, we show the proper "There are no upcoming events" message.

Spotted while reviewing the code, I don't think we've ever actually hit this issue.

### DIFF
--- a/index.html
+++ b/index.html
@@ -45,7 +45,7 @@ layout: home
           {% assign any_events = false %}
           {% for event in site.events %}
           {% comment %}{# Changing this condition? also change it in the below above #}{% endcomment %}
-          {% if event.date > site.time %}
+          {% if event.date > site.time and event.cancelled != true %}
           {% assign any_events = true %}
           {% endif %}
           {% endfor %}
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -44,14 +44,16 @@
           <h3 class="card-heading">Upcoming Events</h3>
           {% assign any_events = false %}
           {% for event in site.events %}
-          {% if event.date > site.time %}
+          {% comment %}{# Changing this condition? also change it in the below above #}{% endcomment %}
```
PeterJCLaw: Yes, I realise this is a bit odd to have two styles of comment. What's going on here is that `liquid` has comments as `{% comment %} ... {% comment %}` ([docs](https://shopify.github.io/liquid/tags/comment/)), but most editors won't do anything special with that.
However, many editors will highlight jinja2 style `{# ... #}` comments, so by putting that inside we get both comment behaviour and visual attention.

sedders123: `in the below above` seems pretty cryptic. Should it just be `in the below` or does the `above` refer to something like position in the rendered page? Either way think this comment needs to be a bit clearer

PeterJCLaw: 8bf09ad46cd854cef02ee61331b300b658b219ce

Issue #162, opened 2019-09-24
Serve the website using GitHub Pages?


RealOrangeOne:
This issue serves more as a discussion point as to whether this is a good idea or not. It's been discussed a couple times, but never went anywhere.

## Currently

The website is served out of a docker container using NGINX's default static file server. This same NGINX configuration also handles reverse proxying for all our other applications ([NGINX config](https://github.com/srobo/website/blob/master/_env/nginx.conf)), whether they be on [`patience`](https://github.com/srobo/server-puppet/tree/patience), or other hosts.

## The problem

Currently, to deploy the website requires rebuilding and re-deploying the entire docker container deployment. Because we don't version our containers, this requires forcing Kubernetes to pull the new container image.

## The solution

As the title suggests, we move the hosting of the website to GitHub Pages, and convert the existing configuration to a simple reverse proxy. Because we use custom plugins, we'll need to build the website on CI, and push it to the `gh-pages` branch manually, rather than rely on GitHub building it for us.

### The benefits

- Auto deployment (we could do this anyway, but it's much simpler)
- Simpler deployment process
- Decouple the Docker configuration and the website (maybe move the docker configuration into [`infrastructure`](https://github.com/srobo/infrastructure/)?)
- GitHub handles the static server configuration rather than us
- Barrier to entry for pushing a change live is much lower

### The drawbacks

- It takes time
- What we have does work, pretty well!
- It's a dependency on a 3rd party
- It's another moving part to consider for the website itself.
- Is there a reason it's setup as it is (historical or otherwise) i'm not considering?
- More people would have access to push changes live (well, by default, we could restrict this if necessary)

@jimmythompson / @PeterJCLaw especially interested in your thoughts.

jimmythompson:
This has been my goal for a while now. Would love a PR. Otherwise I’ll work
on it when I have time.

RealOrangeOne:
Glad to see you share my thoughts. I'll try and throw a PR together in the short term and see how far I can get!

PeterJCLaw:
👍 here.

I suggest for the immediate term that we keep the nginx in this repo _as well_ and have a single CI build which does both. That keeps the moving parts to a minimum. We can then separately extract the `_env` folder (probably to its own repo, though https://github.com/srobo/infrastructure/ might also fit), preserving the history.
[...]

PR #185, opened 2019-12-01
Switch to using GitHub pages

PeterJCLaw:
Changes
---------

This fixes some outstanding issues with regards to our GitHub pages setup (notably around the way that the Jekyll configs are used) and then changes our proxy configuration to point at GitHub pages for the root domain rather than using a local site build.

We still build the site on CI and still have the ability to build it locally, though the CI-built version isn't used anywhere (and doesn't end up in the resulting container).

This essentially replaces https://github.com/srobo/website/pull/165 as most of the changes from that PR are now already in `master` anyway and includes some additional tweaks to the way that the proxy setup is done.

Fixes https://github.com/srobo/website/issues/162.

Testing
-------

I've manually checked that the resulting image runs and serves up a copy of the main website which works, including that 404 pages show as expected for the website.

Follow-on
----------

Once we're happy with these changes I'll duplicate them over into https://github.com/srobo/reverse-proxy and we can look at moving to that instance of the proxy.


_env/nginx.conf
```diff
@@ -173,8 +173,21 @@ http {
...
-      try_files       $uri.html $uri $uri/ =404;
-      index           index.html index.htm;
+      proxy_pass       https://srobo.github.io/website/;
+      proxy_set_header Host srobo.github.io;
+      # Work around github.io redirect issue (see above)
+      proxy_redirect   http://srobo.github.io/website/ /;
+      proxy_redirect   https://srobo.github.io/website/ /;
+
+      sub_filter "/website/" "/";
+      sub_filter_once off;
```
sedders123: Do we definitely want this off? From my (very limited) understanding and reading the [docs](http://nginx.org/en/docs/http/ngx_http_sub_module.html#sub_filter_once) this would replace every instance of `/website/` with a `/` in a URL. It seems unlikely we'd want a sub directory on the site called `/website/` but if we did I believe this setting would break this. Is this desired?

PeterJCLaw: My interpretation is that without this the replacement would only be done one _for the whole document_. I've not tested that and would be happy to drop this if that's not the case though.

PeterJCLaw: Just tested this locally, it does seem to be the case that `sub_filter_once on` means replacing one instance in the whole document while `sub_filter_once off` means doing all replacements.

Just for clarity here, there is an alternative option where we set the underlying Jekyll generation to remove the `/website/` portion, however that would mean that anyone who ended up at the GitHub Pages site would find it rather broken. Since we do currently have bugs which mean that that can happen, I figured it was worth the trade-off for now.

sedders123: 👍 Thanks for looking into this

PR #156, opened 2019-08-31
Add SR2020 signup link

RealOrangeOne:
To be sent out alongside https://github.com/srobo/team-emails/pull/24.

PR #189, opened 2020-01-11
Restore team leader mailing list

RealOrangeOne:
Reverts https://github.com/srobo/website/pull/156 manually, as it wasn't merged using a merge commit.

### DIFF
--- a/_includes/team-signup.html
+++ b/_includes/team-signup.html
@@ -1,0 +1,46 @@
+<form action="//studentrobotics.us12.list-manage.com/subscribe/post?u=1af17c2a3426de258620aae46&amp;id=dd7d04139d"
+      method="post"
+      id="mc-embedded-subscribe-form"
+      name="mc-embedded-subscribe-form"
+      class="form form-left show-labels validate"
+      target="_blank"
+      novalidate>
+    <div id="mc_embed_signup_scroll">
+      <div class="mc-field-group">
+        <label for="mce-EMAIL">Email Address</label>
+        <input type="email" value="" name="EMAIL" class="required email full-width" id="mce-EMAIL">
+      </div>
+      <div class="mc-field-group">
+        <label for="mce-FNAME">Your First Name</label>
+        <input type="text" value="" name="FNAME" class="" id="mce-FNAME" size="24">
+      </div>
+      <div class="mc-field-group">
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
Huh, I had thought we were trying to move away from Mail Chimp?

RealOrangeOne:
We moved away for this year's competition list, namely as a trial to see if it works for our needs.

For now, the list lives in mailchimp. At the end of this competition year I plan to fully review both and perhaps migrate if google groups is suitable (so far, my opinion is that it's not)

For now, I think we should restore the mailchimp list, until we're reached a concrete decision on the list's future.

compete.html
```diff
@@ -130,12 +130,12 @@ <h3>Building a robot</h3>
       <a name="signup"></a><h2>Want to compete?</h2>
       <div class="column m-8-12 m-offset-2-12 text-left">
         <p>
-          Please register your interest below. We'll let you know once places are confirmed. We expect to confirm places at the end of September 2019.
```
PeterJCLaw: lol

_includes/team-signup.html
```diff
@@ -0,0 +1,46 @@
...
+        <a class="button" href="mailto:{{ site.emails.teams }}" target="_blank">Speak to someone</a>
+      </div>
+    </div>
+    <div id="mergeRow-gdpr" class="mergeRow gdpr-mergeRow content__gdprBlock mc-field-group">
+        <div class="content__gdpr">
+            <p>Student Robotics will use the information you provide on this form to contact you around leading a team at Student Robotics. We will contact you via email.</p>
+            <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="gdpr[24613]" tabindex="-1" value="Y"></div>
+        </div>
+        <div class="content__gdprLegal">
+            <p>We use MailChimp as our contacting platform. By clicking below to sign up, you acknowledge that your information will be transferred to MailChimp for processing. Learn more about MailChimp's privacy practices <a href="https://mailchimp.com/legal/" target="_blank">here</a>.</p>
```
PeterJCLaw: Should this also mention that we use Google too?

RealOrangeOne: We don't process mailing list contact using Google unless they sign up for a competition, at which point we state we use Google for this.

So I _think_ under that, we're at least telling the truth!

PR #192, opened 2020-01-23
Serve the runbook from `/runbook/`

RealOrangeOne:
Enabled by https://github.com/srobo/runbook/pull/75

Replaces https://github.com/srobo/infrastructure/pull/6

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -152,7 +152,6 @@ http {
       proxy_set_header Host srobo.github.io;
     }
     
-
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
Links are currently served from `/runbook` anyway, so theoretically there won't be issues there. There's a PR which changes the runbook domain to deal with any domain-absolute paths, on the off chance there are any. I plan to run a dead link checker over the site anyway as part of this to check it has worked properly.

I'll get to deploying this later today.

PR #198, opened 2020-03-11
More competition information

PeterJCLaw:
* Add initial day schedules based on an amalgamation of the schedules in https://github.com/srobo/tasks/issues/436#issuecomment-592980117, with a slight bias towards quoting the earlier times for most things
* Add information about competition parking food & accommodation

We can definitely make these prettier, however the aim here is to get this out quickly now we've realise that the page is as bare as it is.

### DIFF
--- a/_events/sr2020/competition.md
+++ b/_events/sr2020/competition.md
@@ -8,7 +8,7 @@ location: Reading University
 
 The Student Robotics Competition marks the ultimate stage of SR2020, showcasing the months of preparation. Our teams will work hard to push their robots to the limit and quickly make repairs between league matches. We will finish the weekend with the finals where the best bots will be named victorious.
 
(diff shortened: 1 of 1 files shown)


_events/sr2020/competition.md
```diff
@@ -14,6 +14,43 @@ More information will be released soon.
 
 The competition will be held from *Saturday 18th April* until *Sunday 19th April*.
 
+### Day 1 - 18th April, 2020
+
+Saturday begins with a briefing on the rules and general struture of the
+weekend, followed by the start of the league matches will start. The league
```
antoinepetty: I think this is a typo

PeterJCLaw: good spot: 19013a1

PR #200, opened 2020-03-11
Add date to covid-19 news post

antoinepetty:
I think moving the date off this page stopped it being shown in the newsfeed


antoinepetty:
Having checked the other posts, it does not need a date

PR #203, opened 2020-03-12
Add list of COVID-19 blockers to page

RealOrangeOne:
Matches emails https://github.com/srobo/volunteer-emails/pull/29 and https://github.com/srobo/team-emails/pull/47

### DIFF
--- a/covid-19.md
+++ b/covid-19.md
@@ -1,7 +1,7 @@
 ---
 layout: news
 title: COVID-19 Updates
-date: 2020-03-11
+date: 2020-03-12
 ---
 
 We will keep this page updated with new information as it becomes available.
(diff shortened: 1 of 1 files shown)


covid-19.md
```diff
@@ -16,6 +16,16 @@ Considering this is an event for minors, we will be taking decisions with an abu
 
 We are in contact with Reading University Students' Union who are continually reviewing the situation, and mitigating the risks associated with Coronavirus.
 
+The only things that will stop us are:
```
PeterJCLaw: Maybe drop the "only" here? While the list is fairly comprehensive, my pendant mind can find lots of things which aren't present on the list which would still stop us from holding the event.

Not sure if I'm being over-pedantic though.

I think it's because I keep reading "The only things that will stop us" almost like it's a challenge being issued. Maybe something like "The things would prevent the competition from running"?

antoinepetty:
I think it would be worth removing the tech day bit from the bottom (I can't tell if that has been already done on mobile)

covid-19.md
```diff
@@ -16,6 +16,16 @@ Considering this is an event for minors, we will be taking decisions with an abu
...
 
+The things which would prevent the competition from running are:
+
+- If the government says we can't/shouldn't proceed
+- If the venue cancels our booking
+- If we can't get enough teams to come
+- If we can't get enough volunteers to come
+- We don't think we can offer a safe event
+
+All of these are being actively reviewed, and steps being taken to mitigate.
```
sedders123: NIT: This read a little weird to me, maybe something like
```suggestion
All of these are being actively reviewed, and steps are being taken to mitigate these where possible.
```

covid-19.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: news
 title: COVID-19 Updates
-date: 2020-03-11
+date: 2020-03-12
```
PeterJCLaw: We should update this to match the date of publication:
```suggestion
date: 2020-03-14
```

PR #213, opened 2020-07-05
Update links on homepage in news article

PeterJCLaw:
This:
* Canonicalises around the term "Sessions" 
* Copies the links to the event pages into the table in the news article and updates its follow-on spiel to account for that

This is a quick way to make the homepage news article have the latest info (without needing a new article).


PeterJCLaw:
> Feels a bit weird backediting our news, but this makes sense since it's the homepage

Yeah, I'm not sure this is a good idea in general. Maybe we need a better approach for quick updates news posts?

PR #216, opened 2020-07-10
Specify BST for times in event descriptions

antoinepetty:
Fixes https://github.com/srobo/tasks/issues/609


PeterJCLaw:
Should the `date` in the frontmatter also have a timezone?

antoinepetty:
> Should the `date` in the frontmatter also have a timezone?

I'm not sure how it works under the hood, I haven't seen us using timezones with events before

PR #235, opened 2020-11-17
Redirect forum to discord docs

RealOrangeOne:
Fixes https://github.com/srobo/tasks/issues/622

This is intentionally a temporary redirect so it won't affect future deployments.

Will be ported into the new https://github.com/srobo/reverse-proxy repo once approved.

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -43,7 +43,8 @@ http {
     }
 
     location /forum/ {
-      proxy_pass https://patience.studentrobotics.org/forum/;
+      # proxy_pass https://patience.studentrobotics.org/forum/;
+      return 307 https://studentrobotics.org/docs/team_admin/discord;
     }
 
     location /trac/ {
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -44,7 +44,7 @@ http {
 
     location /forum/ {
       # proxy_pass https://patience.studentrobotics.org/forum/;
-      return 307 https://studentrobotics.org/docs/team_admin/discord;
+      return 307 /docs/team_admin/discord;
     }
 
     location /trac/ {


Adimote (dismissed):
LGTM, (I didn't verify it works though, but it agrees with the nginx_http_rewrite_module docs)

PeterJCLaw:
While not a result of this change, unfortunately I get the following error when running this:
```
$ docker run --rm -p 80:80 --name srobo srobo/website
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2020/11/17 21:52:26 [emerg] 1#1: host not found in upstream "saffron.studentrobotics.org" in /etc/nginx/nginx.conf:51
nginx: [emerg] host not found in upstream "saffron.studentrobotics.org" in /etc/nginx/nginx.conf:51
```

I think we need to fix (separately) that before we can merge this.

RealOrangeOne:
It appears this error is appearing even when removing `saffron.sr`. After I do that, it fails on `patience.sr`, which makes very little sense.

It definitely resolves DNS, and the internet is very little help on this.

PeterJCLaw:
`patience` doesn't fail for me, though `srcomp` does. My PRs #237 and #238 fix that.

_env/nginx.conf
```diff
@@ -43,7 +43,8 @@ http {
     }
 
     location /forum/ {
-      proxy_pass https://patience.studentrobotics.org/forum/;
+      # proxy_pass https://patience.studentrobotics.org/forum/;
+      return 307 https://studentrobotics.org/docs/team_admin/discord;
```
PeterJCLaw, line 47: Suggest doing this instead, to avoid any hint of recursion (and also to allow the underlying target to more easily be changed):
```suggestion
      return 307 /docs/team_admin/discord;
```

PeterJCLaw (approved):

PR #240, opened 2020-11-21
Add SR2021 League and Knockout events

sedders123:
Will close https://github.com/srobo/tasks/issues/642

Note: to test the ICS changes here you'll likely need to manually exclude the events which don't have times (or apply https://github.com/srobo/website/pull/243 locally)

PR #242, opened 2020-11-21
Localise event times

PeterJCLaw:
This adds localisation of event times as shown on the website.

Broadly this:
- adds timezones to our source data
- extracts common date display handling
- adds `<time>` elements for machine-readability
- adds some JS to display times in the local timezone
- includes the timezone in the calendar output

### DIFF
--- a/js/localise-times.js
+++ b/js/localise-times.js
@@ -31,7 +31,7 @@ function localise_times(site_timezone) {
 
     let parts = formatter.formatToParts(date).map(({ type, value }) => {
       if (type == 'timeZoneName') {
-        value = '<small style="opacity: 70%;">' + value + "</small>";
+        value = '<small class="muted">' + value + "</small>";
       }
       return { type, value };
     });
--- a/js/localise-times.js
+++ b/js/localise-times.js
@@ -7,6 +7,7 @@ function localise_times(site_timezone) {
     hour12: true,
     minute: 'numeric',
     dayPeriod: 'short',
+    timeZoneName: 'short',
   };
   const date_formatter_with_minutes = Intl.DateTimeFormat(undefined, options);
   delete options.minute;
@@ -28,7 +29,12 @@ function localise_times(site_timezone) {
         ? date_formatter_with_minutes
         : date_formatter_without_minutes;
 
-    let parts = formatter.formatToParts(date);
+    let parts = formatter.formatToParts(date).map(({ type, value }) => {
+      if (type == 'timeZoneName') {
+        value = '<small style="opacity: 70%;">' + value + "</small>";
+      }
+      return { type, value };
+    });
 
     if (navigator.language.startsWith('en-')) {
       // For english speaking visitors where the local format is broadly the
@@ -46,6 +52,6 @@ function localise_times(site_timezone) {
       });
     }
 
-    elem.textContent = parts.map(({ value }) => value).join("");
+    elem.innerHTML = parts.map(({ value }) => value).join("");
   });
 }
--- a/js/localise-times.js
+++ b/js/localise-times.js
@@ -17,7 +17,7 @@ function localise_times(site_timezone) {
   }
 
   document.querySelectorAll("time.time-span").forEach((elem) => {
-    const datetime = elem.parentElement.attributes.datetime.value;
+    const datetime = elem.attributes.datetime.value;
     if (!datetime) {
       return;
     }
--- a/js/localise-times.js
+++ b/js/localise-times.js
@@ -1,0 +1,51 @@
+function localise_times(site_timezone) {
+  let options = {
+    day: 'numeric',
+    month: 'long',
+    year: 'numeric',
+    hour: 'numeric',
+    hour12: true,
+    minute: 'numeric',
+    dayPeriod: 'short',
+  };
+  const date_formatter_with_minutes = Intl.DateTimeFormat(undefined, options);
+  delete options.minute;
+  const date_formatter_without_minutes = Intl.DateTimeFormat(undefined, options);
+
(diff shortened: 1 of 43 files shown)


PeterJCLaw:
-> Draft as I'm going to rebase this on at least https://github.com/srobo/website/pull/240

PeterJCLaw:
Would be great if someone in a different timezone could validate that this does indeed correctly convert the times to their locale.

js/localise-times.js
```diff
@@ -0,0 +1,51 @@
...
+  const date_formatter_with_minutes = Intl.DateTimeFormat(undefined, options);
+  delete options.minute;
+  const date_formatter_without_minutes = Intl.DateTimeFormat(undefined, options);
+
+  if (date_formatter_with_minutes.resolvedOptions().timeZone == site_timezone) {
+    return;
+  }
+
+  document.querySelectorAll("time.time-span").forEach((elem) => {
+    const datetime = elem.parentElement.attributes.datetime.value;
```
sedders123, line 20:
```suggestion
    const datetime = elem.attributes.datetime.value;
```

Without this change I get the following errors in Chrome and Firefox.

image
image

sedders123 (changes requested):
Looks good! Tested locally by changing my system's time zone and the times updated as expected. Did have to make the below change to get things to work though.

sedders123:
Also is it worth stating somewhere clearly that we localise times, or maybe display the timezone alongside the time?

For those viewing the site from outside the UK it seems like they could assume the times specified are UK times given we're based there.

js/localise-times.js
PeterJCLaw: D'oh. This is a result of some refactoring I did.

PeterJCLaw:
> Also is it worth stating somewhere clearly that we localise times, or maybe display the timezone alongside the time?
> 
> For those viewing the site from outside the UK it seems like they could assume the times specified are UK times given we're based there.

Yeah, I'd wondered about this. It's part of why I put the full original time (including timezone) into the `title`. Perhaps that's not obvious enough :thinking:. Part of the challenge is that the timezone names available in JS are long form (e.g: `Europe/London`) rather than, say, `BST`.
We would also only want messaging about localisation to show if the times have in fact been localised (the JS relies on some relatively modern APIs, which may not be present everywhere). That is doable though.

sedders123:
>It's part of why I put the full original time (including timezone) into the title. Perhaps that's not obvious enough 🤔.

Yeah might be ok on desktops (though still not immediately obvious) but I don't think this would work on mobile 😢

Could we do something like the following to get an abbreviation? ([Source](https://stackoverflow.com/a/38708623/5179209))
```js 
const timeZone = /.*\s(.+)/.exec(new Date().toLocaleDateString(navigator.language, { timeZoneName:'short' }))[1];
```

Definitely a fiddly issue!

PeterJCLaw:
> ```js
> const timeZone = /.*\s(.+)/.exec(new Date().toLocaleDateString(navigator.language, { timeZoneName:'short' }))[1]; // "GMT"
> ```

Ewww. However that does provide an option -- I'd tried to work out how to include the timezone in the format I was using, but couldn't. `timeZoneName` is the bit which was missing.

PeterJCLaw:
> ```js
> const timeZone = /.*\s(.+)/.exec(new Date().toLocaleDateString(navigator.language, { timeZoneName:'short' }))[1]; // "GMT"
> ```

I don't think this works like you'd hoped? In Firefox I get `"GMT-5"` if I set my timezone to New York.

PeterJCLaw:
Incidentally, the following avoids the regex:
```js
Intl.DateTimeFormat(undefined, {timeZoneName:'short'}).formatToParts(new Date())[6]
```

sedders123:
>Incidentally, the following avoids the regex:
>
>```js
>Intl.DateTimeFormat(undefined, {timeZoneName:'short'}).formatToParts(new Date())[6]
>```

That's much nicer!

> I don't think this works like you'd hoped? In Firefox I get `"GMT-5"` if I set my timezone to New York.

Its not fantastic but [this SO answer](https://stackoverflow.com/a/61109337/5179209) explains why it works the way it does (TL;DR; Unicode CLDR doesn't contain all the info we need). I think its the best we can get with the inbuilt browser APIs at this point in time. (Even in the shiny [new Temporal proposal](https://tc39.es/proposal-temporal/docs/timezone.html) it looks like there not planning on supporting this 😢.)

We could look add adding a dependency on something like [moment-timezone](https://momentjs.com/timezone/docs/) if we want a wider range of abbreviations, rather than just a few abbreviations with offset fall backs.

sedders123:
Another (much more brittle) option could be we make our own mapping of offsets to abbreviations and populate it for time zones we know that we have teams in, falling back to the "short" timeZoneName for people outside these zones.

Also we may want to use a find (or similar) as I'm not 100% sure that the order or number of items in that list will be the same across different browser implementations.
```js
Intl.DateTimeFormat(undefined, {timeZoneName:'short'}).formatToParts(new Date()).find(p => p.type === 'timeZoneName')
```

PeterJCLaw:
> Also we may want to use a `find` (or similar) as I'm not 100% sure that the order or number of items in that list will be the same across different browser implementations.

The ordering will be defined by the locale, so it will definitely vary. What I'd expect to do is to modify the existing usage of `DateTimeFormat` so that it includes a `timeZoneName` component and just let that put it where it wants. I'd be tempted to do something like surrounding the component in `<small style="opacity: 0.7">` or similar to de-emphasize it a bit though.

sedders123 (approved):
This looks good to me! Tested in a few different time zones some with abbreviations and some without.

PR #243, opened 2020-11-22
Exclude events without a datetime set

PeterJCLaw:
This is necessary to ensure that our calendar file is valid.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,8 @@ jobs:
       - run: docker info
       - run: rake build
 
+      - run: rake validate
+
   build_and_release:
     docker: *docker-template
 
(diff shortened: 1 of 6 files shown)


sedders123 (dismissed):
Tested this locally and I can see that the calendar file now excludes events without a time specified 🎉

I tried to run the validate script locally but got the below error. Thinks its probably cause I'm running on Windows. I did `chmod +x` the file but it didn't seem to help. As the scripts is run on CI I'm not too worried about it

image

sedders123:
Ahh found it, on Windows I need to run `bundle exec ruby scripts/validate-icalendar.rb` as the shebang isn't picked up!

sedders123 (approved):

PR #259, opened 2021-02-07
Update to a more recent Ruby version

PeterJCLaw:
This reflects the likely version users have to hand.

Hopefully this also increase the chances the image is already cached on the CI host.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build_only:
     docker: &docker-template
-    - image: circleci/ruby:2.5
+    - image: circleci/ruby:2.7
 
     steps:
       - checkout


RealOrangeOne:
This output spits out a lot of warnings in the console. It'd probably be worth doing a Jekyll upgrade at the same time.

PeterJCLaw:
> This output spits out a lot of warnings in the console. It'd probably be worth doing a Jekyll upgrade at the same time.

It does, however ruby 2.7 is the version which Github Pages is using and many people (myself included) are already using it locally. I agree that we should also fix the warnings, though really we should probably move to Bundle 2 as well... this starts to get quite involved. I'd rather get this one in and then handle those parts separately.

RealOrangeOne (approved):
I too am using 2.7 locally, and know this to be working.

PR #263, opened 2021-03-20
Set the timings for the SR2021 knockouts

PeterJCLaw:
Fixes https://github.com/srobo/tasks/issues/643.
Timings per previous discussions.
Would be great to also get the youtube livestream link in here too, though this is likely useful on its own.


RealOrangeOne:
@PeterJCLaw livestream link will be https://youtu.be/gxJM6ernMqo

RealOrangeOne (approved):

PR #264, opened 2021-04-03
Update dependencies

PeterJCLaw:
This aims to minimally update our dependencies for a Kramdown security update.

From manual diffing of the resulting site build, there is a minimal change here -- just some extra `role` attributes on the generated elements (mostly footnotes).


PeterJCLaw:
In case it helps, we're almost certainly already using these versions in production -- here's the versions which GitHub Pages uses: https://pages.github.com/versions/

sedders123 (approved):

PR #270, opened 2021-05-01
Disable competition mode

PeterJCLaw:


RealOrangeOne:
The diff here doesn't quite look right - in fact there isn't one.

PeterJCLaw:
> The diff here doesn't quite look right - in fact there isn't one.

That's because this PR contains both the enable and disable commits due to having been created before the enable commit went live. The lack of diff is therefore expected.

RealOrangeOne (approved):

PR #273, opened 2021-05-01
Update simulator link

PeterJCLaw:
While we left a redirect in place it seems that that only works for the path without a trailing slash. This change updates the url to the new one and removes the trailing slash so that it should continue to work even if the url changes again.

### DIFF
--- a/_posts/2020-07-25-post-competition.md
+++ b/_posts/2020-07-25-post-competition.md
@@ -145,7 +145,7 @@ If you would like to find out more, please [get in touch][contactus].
 
 _The SR Team_
 
-[simulator]: https://studentrobotics.org/docs/competition-simulator/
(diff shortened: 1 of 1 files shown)


antoinepetty (approved):

antoinepetty (approved):

antoinepetty:
Doesn't seem to like my approval - I am not worthy

PeterJCLaw:
@RealOrangeOne do you know why @antoinepetty can't approve this/

RealOrangeOne:
I don't have access to see the permissions setup, but my assumption would be that it's set up so that volunteers not in a "team" don't have write access to the repo (required for an approval to matter)

RealOrangeOne (approved):

PR #274, opened 2021-07-08
Optimise images

raccube:
Used ImageOptim with default settings (lossless)

Saved 1,1 MB out of 7,8 MB. 15,1 % per file on average (up to 46,8 %)


sedders123:
Can we add something to automate this in the future?

RealOrangeOne:
Having a helper script to optimise an image would be good. Optimising an already optimised image is probably not a good idea. Even if only so we can point to it when people are uploading images in future.

sedders123 (dismissed):

raccube:
Could definitely make a rake task for that but it would potentially, if we don't use lossy compression most image optimisers wouldn't do anything if they deem an image as already optimised.

RealOrangeOne:
Having a pre-built rake task with a system dependency feels fine. Can easily add some install instructions into the README

PeterJCLaw:
Impressive savings. We've definitely used https://tinypng.com/ on images before, so I'm actually surprised how much we're able to compress them further :)

raccube:
Some images were skipped as they took too long. I can try and rerun those manually on a more powerful machine.

raccube:
That's all I could do.

sedders123 (approved):

PR #276, opened 2021-07-14
Remove IDE from navbar

RealOrangeOne:
We don't really need it there

image

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -6,9 +6,6 @@
   </div>
   <ul id="navigation" class="full-navigation mobile-collapsed">
     <li>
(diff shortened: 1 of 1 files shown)


raccube (approved):

PeterJCLaw:
Currently the nav bar is consistent between the main website, the docs and the competition website. What's the plan for the other areas of the site?

I mostly agree that the IDE doesn't need to be in the main site nav bar (especially as we move towards less dependency on it), however I think we need to consider the broader scope and what divergence we expect from different parts of the site.

antoinepetty:
Covered by https://github.com/srobo/website/pull/303

PR #277, opened 2021-07-14
Add volunteer testimonials

RealOrangeOne:
These were obtained from the 2019 competitor booklet (https://github.com/srobo/competitor-booklet-2019/releases/tag/SR2019) and the dates updated

### DIFF
--- a/_sass/base.scss
+++ b/_sass/base.scss
@@ -51,3 +51,7 @@ img {
   width: 100%;
   height: auto;
 }
+
+.text-italic {
+  font-style: italic;
+}
--- a/volunteer.html
+++ b/volunteer.html
@@ -69,6 +69,44 @@ permalink: /volunteer/
 </div>
 <div class="panel">
   <div class="panel-container">
+    <div class="fixed-width">
+      <h2 class="text-center">Testimonials</h2>
+      <div class="row">
+        <div class="column s-4-12">
+          <blockquote class="text-italic">
+            After 1 kickstart event, I was hooked, and haven’t looked back since! After finishing college, I just had to become a blueshirt, to help those like me realise their development potential too.
+          </blockquote>
+          <p class="text-heavy text-center">- Jake Howard</p>
+          <h5 class="text-center">
+            Team CLY SR2014-15<br>
+            Blue Shirt SR2016-21
+          </h5>
+        </div>
+        <div class="column s-4-12">
+          <blockquote class="text-italic">
+            SR was the main reason I signed up to do electronics at university, the experience gained by mentoring and coaching on tech days has been invaluable.
+          </blockquote>
+          <p class="text-heavy text-center">- Tom Wheal</p>
+          <h5 class="text-center">
+            Team PSC SR2014-15<br>
+            Blue Shirt SR2017-21
+          </h5>
+        </div>
+        <div class="column s-4-12">
+          <blockquote class="text-italic">
+            Volunteering with Student Robotics gave me the opportunity to work on projects with real deadlines. When applying for my first professional jobs, that experience let me stand out from other applicants.
+          </blockquote>
+          <p class="text-heavy text-center">- Peter Law</p>
+          <h5 class="text-center">
+            Blue Shirt SR2007-21
+          </h5>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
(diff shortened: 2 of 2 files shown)


volunteer.html
```diff
@@ -67,6 +67,44 @@ <h3 class="text-center">Development</h3>
...
+          </blockquote>
+          <p class="text-heavy text-center">- Tom Wheal</p>
+          <h5 class="text-center">
+            Team PSC SR2014-15<br>
+            Blue Shirt SR2017-21
+          </h5>
+        </div>
+        <div class="column s-4-12">
+          <blockquote class="text-italic">
+            Volunteering with Student Robotics gave me the opportunity to work on projects with real deadlines. When applying for my first professional jobs, that experience let me stand out from other applicants.
```
PeterJCLaw, line 97: I wrote this? (It's accurate, I just don't remember)
It also feels (to me) a bit dated (the context described is about a decade old). I'm not sure what I'd say instead though. Will have a think and see if there's something I can suggest instead.

volunteer.html
```diff
@@ -67,6 +67,44 @@ <h3 class="text-center">Development</h3>
...
+          </blockquote>
+          <p class="text-heavy text-center">- Peter Law</p>
```
PeterJCLaw, line 99: Should we use longer dashes here?

volunteer.html
```diff
@@ -67,6 +67,44 @@ <h3 class="text-center">Development</h3>
...
+          <h5 class="text-center">
+            Blue Shirt SR2007-21
```
PeterJCLaw, line 101: Is "Blue Shirt" the term we want to use public-facing? It has a lot of context for us (and when you're actually at an event), but it's not the clearest term.
I realise we do use it elsewhere on the page (though as a single word, not two), though maybe we should reconsider that too?

Separately is using SR years for volunteer timeframes really useful? Again they don't feel entirely clear to the public and they don't really work well for people (such as myself) whose volunteering timeframes don't line up with the competition years. I suggest using calendar years instead.

Finally, including an end date on these kinda implies that the people mentioned have stopped being volunteers. Perhaps something like "since 2006" would be clearer?

antoinepetty: I've changed the captions underneath to be "volunteer" but have kept the term "blueshirt" in one testimonial and an instance above. I think it's a good term to keep if we can make it clear what's meant somewhere on the page

antoinepetty (approved):

volunteer.html
PeterJCLaw: https://github.com/srobo/website/pull/277/commits/0429890e1fca01e4dbe21f6bfb753e329abb2c71

volunteer.html
PeterJCLaw: https://github.com/srobo/website/pull/277/commits/0429890e1fca01e4dbe21f6bfb753e329abb2c71

_sass/base.scss
```diff
@@ -51,3 +51,7 @@ img {
   width: 100%;
   height: auto;
 }
+
+.text-italic {
```
PeterJCLaw, line 55: Suggest that we prefer to use `<em>` in most cases and then sparingly apply style semantically rather than functionally. For this case having a `quotation` class (which applies this styling) might make sense? Makes it more obvious when adding a new quotation that it should have the same style.

sedders123:
Its a shame the three testimonials given here are all from our male volunteers. To encourage a more diverse pool of potential volunteers to join us I think we should try and get a few testimonials from volunteers of other genders.

I like the three quotes displayed next to each other from a design perspective so maybe we could change to a carousel to allow for additional testimonials to be added over time?

antoinepetty:
@sedders123 personally I see this as a stop-gap solution which we could revisit after freshening up the homepage for sponsors? I agree a carousel and quotes from more diverse volunteers (both in terms of demographics and career paths) would be better.

sedders123:
Do we need a stop gap or can we just do it properly in one go? Recruiting new volunteers is essential but until we've got the other processes in place to handle them once they've signed up, from my perspective I don't think this PR is urgent. More then happy to be convinced otherwise though :)

antoinepetty:
I don't feel strongly so am happy to wait until we have something better before deploying. Little profile photos would be a nice addition too

PR #278, opened 2021-07-14
Upgrade to Bundler 2

raccube:


RealOrangeOne (changes requested):
Besides some local-specific "qwerks", looks like this is working fine. Nice to not have to do the bundler dance to setup!

Looks like the bundler version is still explicitly set in `.circleci/config.yml`

README.md
```diff
@@ -6,12 +6,12 @@ The Student Robotics public website.
 
 ## Getting Started
 
-1. [Install Ruby][install-ruby]
+1. [Install Ruby 2.7][install-ruby]
```
PeterJCLaw, line 9: Do we care that it's 2.7 specifically, or would a later version be ok?
(Might be useful to record somewhere, though perhaps not in the setup instructions) why a specific version is needed)

PeterJCLaw:
Nice one, thanks for starting on this (I've been hoping to do this for ages).

If you haven't already done so, I suggest getting people on a variety of platforms (ideally covering at least Windows/macOS/Linux) to check that the setup instructions work for them. In the past I've found that Ruby setup on different platforms can be quite finicky unfortunately.

README.md
raccube: [GitHub Pages uses 2.7 specifically.](https://pages.github.com/versions/) We probably could use 3.0 but I think it makes sense to stick to what is used on GitHub Pages.

RealOrangeOne:
>  people on a variety of platforms [...] to check

Perhaps a pre-requisite to this should be moving to GitHub actions, where we can have computers do this testing for us

PeterJCLaw:
> > people on a variety of platforms [...] to check
> 
> Perhaps a pre-requisite to this should be moving to GitHub actions, where we can have computers do this testing for us

I'm not sure that helps? The install approach used in CI is inevitably going to differ to that we indicate users do in the README. Things like "check this checkbox in the installer" aren't think which CI can catch.

raccube:
I just tested this on Windows running a native version of Ruby and it runs fine.

sedders123 (approved):

PR #280, opened 2021-07-16
Add LinkedIn and Instagram to footer

### LINKED ISSUE #210: Add instagram link and icon to footer
Currently have links to our other socials but not instagram 
image

https://www.instagram.com/student_robotics/

raccube:
- Add title to the social media icons in the footer
- Update Font Awesome to 4.7 so that we have an up to date Instagram icon

Closes #210 

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -43,7 +43,7 @@ social:
     icon: twitter
     link: https://twitter.com/studentrobotics
   - title: YouTube
-    icon: youtube
+    icon: youtube-play
     link: https://youtube.com/user/StudentRobotics
   - title: Instagram
     icon: instagram
--- a/_config.yml
+++ b/_config.yml
@@ -36,10 +36,24 @@ emails:
   info: info@studentrobotics.org
 
 social:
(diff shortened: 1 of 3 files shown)


RealOrangeOne:
I feel like we need need these icons in decreasing order of importance, rather than increasing (read: big-endian vs little-endian). 

GItHub, Twitter and YouTube are all highly used by us, Instagram is also fairly regularly used. Facebook and LinkedIn on the other hand, less so. 

Alternatively, alphabetical order might be better?

RealOrangeOne (dismissed):

PeterJCLaw (approved):
One gotcha to note here is that LinkedIn seems to _require_ you to sign up to see our page.
If there's a setting we can tweak which makes it public we should perhaps do that? Don't think that's a blocker to this PR though.

RealOrangeOne:
(As noted in #276, this should be ported to the docs and competition-website too)

raccube:
Can someone merge this?

PR #282, opened 2021-07-31
Add donation buttons to nav bar and homepage

sedders123:
image

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -17,5 +17,8 @@
     <li>
       <a class="page-link" href="{{ '/volunteer' | prepend: site.baseurl }}">Volunteer</a>
     </li>
+    <li>
+      <a class="page-link" href="https://link.justgiving.com/v1/charity/donate/charityId/3118662?tipScheme=TipJar2.1&isRecurring=true&amount=20.00&reference=givingcheckout_tj21">Donate</a>
+    </li>
   </ul>
 </nav>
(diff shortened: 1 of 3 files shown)


RealOrangeOne:
I notice the "Donate" box sits above "Sponsors" - is that intentional? I feel like it should probably be below.

sedders123:
> I notice the "Donate" box sits above "Sponsors" - is that intentional? I feel like it should probably be below.

Have responded in slack https://studentrobotics.slack.com/archives/CCD59522G/p1627730696000300?thread_ts=1627730370.000100&cid=CCD59522G

RealOrangeOne (dismissed):

_includes/navigation.html
```diff
@@ -17,5 +17,8 @@
     <li>
       <a class="page-link" href="{{ '/volunteer' | prepend: site.baseurl }}">Volunteer</a>
     </li>
+    <li>
+      <a class="page-link" href="https://link.justgiving.com/v1/charity/donate/charityId/3118662?tipScheme=TipJar2.1&isRecurring=true&amount=20.00&reference=givingcheckout_tj21">Donate</a>
```
PeterJCLaw, line 21: Suggest pulling the url out to a configuration value, not because it'll change but so that we have a single constant reference rather than lots of literals. This will also make it easier to copy across into our sub-sites if we want.

There might also be an argument for having a redirect from, say, `/donate`, though that might be beyond what we need for now.

Scarzy: I can actually see examples where it might change, e.g. if we decide to tweak the default values. Which would support the idea of having it in a variable.

/donate also sounds like a good idea, but like you say a separate task

sedders123: Done in 3c8899a :)

RealOrangeOne (approved):

Scarzy (approved):

PR #293, opened 2021-09-04
Add support for social media meta tags

sedders123:
Adds OpenGraph and Twitter meta tags so our links can be displayed nicely on social media platforms


_includes/head.html
```diff
@@ -1,13 +1,48 @@
...
+{% else %}
+  {% assign description = site.description %}
+{% endif %}
 <head>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
-  <title>{% if page.title %}{{ page.title }} | {% endif %}{{ site.title }}</title>
-  <meta name="description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description }}{% endif %}">
```
PeterJCLaw, line 7: Seems like we're dropping support for `excerpt` (which I think is auto generated for blog posts?) -- is that intentional?

_includes/head.html
```diff
@@ -1,13 +1,48 @@
...
+  <meta property="og:locale" content="{{ site.locale }}">
+
+  {% if page.image %}
+    <meta property="og:image" content="{{ page.image | absolute_url }}">
+    {% if page.image_alt %}
+      <meta property="og:image:alt" content="{{ page.image_alt }}">
+    {% endif %}
+    <meta name="twitter:card" content="summary_large_image">
+  {% else %}
+    <meta property="og:image" content="{{ 'images/logo-image-only.png' | absolute_url }}">
```
PeterJCLaw, line 35: Is there support for using an SVG here?

sedders123: [Unfortunately it seems we can't use an SVG](https://indieweb.org/The-Open-Graph-protocol#Does_not_support_SVG_images)

_includes/head.html
sedders123: Ahh was unaware `excerpt` was autogenerated, didn't find it anywhere in the code base so had assumed it was unused. Have added support for this back in, if the description is not explicitly overridden. 069a8c8

_includes/head.html
```diff
@@ -1,13 +1,50 @@
+{% if page.title %}
+  {% assign title = page.title %}
```
RealOrangeOne, line 2: This changes the titles of pages to remove the `| Student Robotics` from pages. Is that intentional?

_includes/head.html
PeterJCLaw: Hrm, seems we may need to opt-in to them being generated: https://jekyllrb.com/docs/pages/#excerpts-for-pages. Feels like it would be a good thing to opt into though!

_includes/head.html
sedders123: Originally removed it to make the OpenGraph title the same as the page title but on further thought there is no reason these need to match so have added back 223b255

_includes/head.html
sedders123: 👍 d13ee2e

PeterJCLaw (approved):
Not sure how best to test this against what'll actually show up on Twitter etc., but seems good from what I can see here.

sedders123:
To test this I ran [`ngrok`](https://ngrok.com/) locally and used https://cards-dev.twitter.com/validator and https://developers.facebook.com/tools/debug/

RealOrangeOne (approved):

PR #295, opened 2021-09-05
Add social media share buttons to posts

sedders123:
Adds share buttons to the sidebar on a post. Can be seen in the below image
image

### DIFF
--- a/_includes/social-share-buttons.html
+++ b/_includes/social-share-buttons.html
@@ -5,8 +5,8 @@
   <span>Share on: </span>
   <div class="social-share-links">
       <a class="facebook" title="Share this on Facebook" href="https://www.facebook.com/sharer.php?u={{ pageurl }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-facebook"></i></a>
-      <a class="twitter" title="Share this on Twitter" href="https://twitter.com/intent/tweet?url={{ pageurl }}&text={{ page.title | url_encode }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-twitter"></i></a>
+      <a class="twitter" title="Share this on Twitter" href="https://twitter.com/intent/tweet?url={{ pageurl }}&text={{ page.title | cgi_escape }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-twitter"></i></a>
       <a class="linkedin" title="Share this on Linkedin" href="https://www.linkedin.com/shareArticle?mini=true&url={{ pageurl }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-linkedin"></i></a>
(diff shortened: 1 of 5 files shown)


RealOrangeOne:
I recently was shown https://shareon.js.org/, which looks like an also quite nice way to handle these share links. Perhaps a little heavier than we need, but makes it easy to support a lot of different platforms.

_includes/social-share-buttons.html
```diff
@@ -0,0 +1,12 @@
+{% assign pageurl = page.url | absolute_url %}
+{% include get_description_for_page.html %}
+
+<div class="social-share">
+  <span>Share on: </span>
+  <div class="social-share-links">
+      <a class="facebook" title="Share this on Facebook" href="https://www.facebook.com/sharer.php?u={{ pageurl }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-facebook"></i></a>
```
PeterJCLaw, line 7: Do the urls not also need `url_encod`ing? We almost certainly don't want `&`s in the post url being interpreted as part of the share target url (etc.).

sedders123: Yes that's a great catch! Though we'd have to `cgi_escape` rather than `url_encode` to ensure we escape ampersands a43c52e67ad8feff48d03fa971db5db285604acb

sedders123:
> I recently was shown https://shareon.js.org/, which looks like an also quite nice way to handle these share links. Perhaps a little heavier than we need, but makes it easy to support a lot of different platforms.

It does look nice, but referencing an external library for what is essentially 4 links feels a little overkill to me

_includes/social-share-buttons.html
PeterJCLaw: Hrm, it feels like this ought to have the same escaping as the text we're pushing in the urls for the template titles. The fact that this is originally a url shouldn't change the encoding scheme being used. I'm not sure which is correct, though it does feel like it ought to be consistent.

raccube:
> > I recently was shown https://shareon.js.org/, which looks like an also quite nice way to handle these share links. Perhaps a little heavier than we need, but makes it easy to support a lot of different platforms.
> 
> It does look nice, but referencing an external library for what is essentially 4 links feels a little overkill to me

There's also the [Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API) which is quite straight forward to implement.

PeterJCLaw:
:cough: I think we should paint the bikeshed blue? :cough: 
image
(I agree with James -- just having the links directly in our templates seems a much better solution than taking a JavaScript dependency)

_includes/social-share-buttons.html
sedders123: After some playing about locally I have come up with 32f56258e77540b8f7b0093523132316f5a18dab. This ensures that ampersands still get sent through correctly and also accounts for `mailto:` not liking `+` in place of `%20`.

antoinepetty (approved):

antoinepetty:
Soooooo the linkedin and facebook buttons don't work for me (sorry I should have tested more thoroughly before approving)

Issue #298, opened 2021-09-11
Create a Teams page


antoinepetty:
A page for team and competition-year specific information. This is essentially our current homepage, but needs moving now that #286 is happening

antoinepetty:
Part of this change should be removing the header links to IDE and Docs, having a Teams option instead

PR #300, opened 2021-09-11
New Sponsor-focused homepage

### LINKED ISSUE #286: Spruce up homepage with focus on sponsors
Needs to be similar to https://studentrobotics.org/about/ but with more photos, testimonials etc

antoinepetty:
Work in progress - essentially combining the best bits of /about and /compete, then aiming them at a general audience

Closes #286 

The old homepage content will be moved to a new page as per #298


antoinepetty:
I changed the banner text from `An annual robotics competition for young people aged 16-18` to `Inspiring the next generation of engineers`
I'm on the fence about which is best for the homepage

antoinepetty:
Current state:
screencapture-127-0-0-1-4000-website-2021-09-11-21_49_09

sedders123:
Added some tweaks to the layout. ([Updated screenshot](https://i.imgur.com/tsbNSql.png)).
Things we still need to do:
 - [ ] Review and update copy where necessary
 - [ ] Add more photos to each section wherever possible
 - [ ] Do we need a call to action in the hero? Feels a little empty at the moment but not sure what should be there

antoinepetty:
Current state:
screencapture-127-0-0-1-4000-website-2021-09-18-20_34_51

raccube:
@antoinepetty Can we get a mobile screenshot?

sedders123:
Here's the [latest screenshots](https://imgur.com/a/ctLbCnk) on both web and mobile

index.html
```diff
@@ -1,47 +1,160 @@
...
+        <h2>Valuable skills for young people</h2>
+        <p>
+          Student Robotics primes participants for good careers in engineering.
+          Participants assemble a team, build a robot, and compete. All the while
+          gaining valuable technical and soft skills to help their future careers.
+        </p>
+      </div>
+      <div class="row">
+        <div class="column center-label s-4-12">
+          <img src="{{ '/images/advertising/happy-team.jpg' | prepend: site.baseurl }}" alt="" />
```
raccube, line 60: Should probably add alt text here

index.html
```diff
@@ -1,47 +1,160 @@
...
 ---
-<div class="main home-main">
-  <div class="card">
+<div class="panel">
+  <div class="panel-container">
     <div class="fixed-width">
+      <h2 class="text-center">What is Student Robotics?</h2>
+      <div class="row vertically-centered-row">
+        <div class="column m-6-12">
+          <img src="{{ '/images/advertising/team-in-arena.jpg' | prepend: site.baseurl }}" />
```
raccube, line 10: Add alt text or empty alt attr

raccube (changes requested):

_layouts/home.html
```diff
@@ -4,18 +4,21 @@
...
+      <img class="logo" src="{{ '/images/logo-text-white.png' | prepend: site.baseurl }}" alt="Student Robotics" />
     </a>
     {% include navigation.html class="site-navigation home-navigation" %}
   </div>
   <div class="hero-container hero-homepage">
     <div class="fixed-width">
       <div class="hero-text">
         <h1>An annual robotics competition for young people aged 16-18</h1>
-        <a class="button button-hero" href="{{ '/about' | prepend: site.baseurl }}">Learn more</a>
+        <a class="button hero-homepage-button" href="{{ '/compete' | prepend: site.baseurl }}">Compete</a>
```
RealOrangeOne, line 15: Question: There's a lot of duplication between these CTAs and the navbar - is that intentional?

_layouts/home.html
```diff
@@ -4,18 +4,21 @@
...
+        <a class="button hero-homepage-button" href="{{ '/volunteer' | prepend: site.baseurl }}">Volunteer</a>
+        <a class="button hero-homepage-button" href="{{ 'http://link.justgiving.com/v1/charity/donate/charityId/3118662?tipScheme=TipJar2.1&isRecurring=true&amount=20.00&reference=givingcheckout_tj21' }}">Donate</a>
```
RealOrangeOne, line 17:
```suggestion
        <a class="button hero-homepage-button" href="http://link.justgiving.com/v1/charity/donate/charityId/3118662?tipScheme=TipJar2.1&isRecurring=true&amount=20.00&reference=givingcheckout_tj21">Donate</a>
```

Additionally, is this not defined in a variable somewhere?

about.html
```diff
@@ -1,109 +0,0 @@
----
```
RealOrangeOne, line 1: Thought: Should we have a redirect for this page?

index.html
```diff
@@ -1,47 +1,160 @@
...
+            src="{{ '/images/advertising/headshots/james-seden-smith.jpg' | prepend: site.baseurl }}" />
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+<div class="panel panel-highlight panel-thin">
+  <div class="panel-container">
+    <div class="fixed-width">
+      <h2 class="text-center top-spacing">Competitor feedback from the last two years</h2>
```
RealOrangeOne, line 122: Suggestion: Should this read more like a sentence, so it flows better with each of the below points?

```suggestion
      <h2 class="text-center top-spacing">Of competitors surveyed from the last 2 years...</h2>
```

index.html
```diff
@@ -1,47 +1,160 @@
...
-    <div class="row">
-      <div class="sidebar column m-4-12">
-        <div class="card">
-          <h3 class="card-heading">Upcoming Events</h3>
+</div>
+<div class="panel">
+  <div class="panel-container">
+    <div class="fixed-width text-center">
+      <div class="row">
+        <div class="column m-8-12">
```
RealOrangeOne, line 144: Question: Is there a reason posts is twice as wide as events? 50:50 might look nicer? Or perhaps use this weighted layout just when there are no events?

index.html
```diff
@@ -1,47 +1,160 @@
...
+<div class="panel">
+  <div class="panel-container">
     <div class="fixed-width">
+      <h2 class="text-center">What is Student Robotics?</h2>
+      <div class="row vertically-centered-row">
+        <div class="column m-6-12">
+          <img src="{{ '/images/advertising/team-in-arena.jpg' | prepend: site.baseurl }}" />
+        </div>
+        <div class="column m-6-12">
+          <p class="about-p">
```
RealOrangeOne, line 13: Thought: Should there be a little more (horizontal) padding between the text and image?

image

sedders123:
Thanks for the feedback so far! We're going to hold off making any changes until the trustees have given their feedback on the design, in case changes are needed. I have emailed screenshots of the redesign to the trustees and will update here with any feedback we get :)

raccube (approved):

index.html
```diff
@@ -1,47 +1,199 @@
...
+        </div>
+        <div class="column m-4-12">
+          <p class="statistic">88%</p>
+          <p class="statistic-label">are now highly interested in a STEM career</p>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+<div class="panel">
```
Scarzy, line 144: Does it make sense to put the code for the sponsors box into a separate file, so it can be easily brought into any page?

index.html
```diff
@@ -1,47 +1,199 @@
...
+          </blockquote>
+          <p class="text-heavy text-center">— James Seden Smith</p>
+          <h5 class="text-center">
+            Student Robotics 2015 and 2016<br>
+            Team CLY
+          </h5>
+        </div>
+        <div class="column m-3-12 horizontally-centered-column">
+          <img class="testomonial-avatar"
+            src="{{ '/images/advertising/headshots/james-seden-smith.jpg' | prepend: site.baseurl }}"
```
Scarzy, line 116: I suggest you come up with a different naming scheme for these images. I assume in this case as the person adding it was also the one pictured they don't mind their full name being public, but we shouldn't assume that this will be the norm, or even if they are happy that we can make this the norm. It will also be easier and give a better feel to the page to be consistent rather than some have a surname and others don't

Perhaps something like "volunteer-james-1.jpg"?

index.html
```diff
@@ -1,47 +1,199 @@
...
+      <div class="row vertically-centered-row">
+        <div class="column m-9-12">
+          <blockquote class="text-italic">
+            Student Robotics provided me with the invaluable experience of working on a engineering project with a team.
+            I learned not only good engineering principals but the importance of being able to clearly communicate my
+            thoughts and reasoning.
+            Competing in Student Robotics opened the door to me directly entering the world of work as a Software
+            Engineer after completing Sixth Form.
+          </blockquote>
+          <p class="text-heavy text-center">— James Seden Smith</p>
```
Scarzy, line 108: See my comment regarding names and consistency on the image name. Same applies here

index.html
sedders123: I feel like if we wanted to show sponsors on other pages we may not want them displayed in the same way as this page (we may want them in a side panel, or in a different orientation for example). I tend to prefer pulling stuff out into a shared file when we use it in another place, as opposed to pre-emptively splitting it out.

index.html
sedders123: Antoine has done this is in 5d2c6fa

index.html
sedders123: Antoine has done this in 5d2c6fa

about.html
sedders123: df3af87

raccube (approved):

raccube:
For the record, I checked it out locally and tested in Firefox on Linux as well as on iOS (iPhone 12).

raccube:
image

Only nitpick would be that this text gets a bit close to the edges, and having a line wrap for just one word looks a bit weird. Otherwise I think it looks good.

PR #303, opened 2021-09-18
New Teams page

antoinepetty:
Moves the existing homepage to a new team-focused page to make way for https://github.com/srobo/website/pull/300

closes https://github.com/srobo/website/issues/298

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -6,10 +6,7 @@
   </div>
   <ul id="navigation" class="full-navigation mobile-collapsed">
     <li>
-      <a class="page-link" href="/ide">IDE</a>
-    </li>
-    <li>
-      <a class="page-link" href="/docs">Documentation</a>
+      <a class="page-link" href="/teams">Teams</a>
     </li>
     <li>
       <a class="page-link" href="{{ '/compete' | prepend: site.baseurl }}">Compete</a>
--- a/teams.html
+++ b/teams.html
@@ -1,0 +1,114 @@
+---
+layout: page
+title: About
+permalink: /teams/
+---
+<div class="panel panel-header panel-boxes">
+  <div class="panel-container">
+    <div class="hero-text no-cta fixed-width text-center">
+      <h1>Teams Dashboard</h1>
+    </div>
+  </div>
+</div>
+<div class="main home-main">
+  <div class="card">
+    <div class="fixed-width">
+      <div class="row">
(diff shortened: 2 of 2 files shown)


antoinepetty:
TODO: change navbar to match on docs, competition website as per https://github.com/srobo/website/pull/276
This pull request supersedes https://github.com/srobo/website/pull/276

_includes/footer.html
```diff
@@ -3,9 +3,9 @@
     <div class="column d-4-12 m-2-12">
       <h4>For Competitors</h4>
       <ul class="sitemap">
+        <li><a href="/teams/">Teams Dashboard</a></li>
```
PeterJCLaw, line 6: (drive-by): I think this needs `site.baseurl` to ensure it works correctly locally:
```suggestion
        <li><a href="{{ '/teams/' | prepend: site.baseurl }}">Teams Dashboard</a></li>
```

antoinepetty:
todo: add site.baseurl to all footer and nav links and ensure they work

PeterJCLaw:
> todo: add site.baseurl to all footer and nav links and ensure they work

Sooo, this is kinda complicated. The pattern I've used is that anything which is in the current repo should have `site.baseurl`, but anything which is on the website via other means shouldn't.
For the latter I've often tended towards using a full url (i.e: including the domain etc.) so that they work correctly locally. That does mean that if you were running them all side-by-side locally they wouldn't cross link, however I've not seen anyone do that and it does mean that it's clearer which urls are fully internal vs external to the current project.

antoinepetty:
Reverted that commit so that the format matches what it's replacing

PeterJCLaw:
@antoinepetty apologies for confusion here regarding the format of links. My expectation is that any links which are internal to this repo should have `site.baseurl` prepended in some form. Doing so ensures that they're functional both in prod and development.
However, links to things which aren't in this repo should not have that prefix added since it would result in weirdness at least in development (and may or not break in prod, not sure). I realise that this means that some of the nav links are in different formats, but that's fine since the format relates to the thing they link to rather than the place they are within our code.

Happy to chat this through if that would help.

(I suspect a style guide around this might help in future; I've created https://github.com/srobo/tasks/issues/700 for that)

antoinepetty:
Ah I see the pattern now, thank you

sedders123 (approved):

PR #306, opened 2021-09-23
Make the sponsor checkmark list markers more like list markers

PeterJCLaw:
**Before**

image

**After**

image


sedders123 (approved):
Thanks Peter, this looks great! (Also checked mobile and all good there)

antoinepetty (approved):

PR #309, opened 2021-09-27
Validate the nginx config on CI

RealOrangeOne:
We should validate that the nginx config is valid during CI. It's much easier to debug here than when it comes to deploying it to production!

See also https://github.com/srobo/reverse-proxy/issues/16


RealOrangeOne:
Appears to be running correctly:

image

PR #307, opened 2021-09-23
Add Adventurous Machines as sponsor

antoinepetty:

PR #311, opened 2021-10-04
Swap order of sponsors

antoinepetty:
Probably best to keep them alphabetical until another order is defined

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -68,13 +68,12 @@ social:
     link: https://www.linkedin.com/company/10969472/
 
 sponsors:
-  - name: University of Southampton
-    link: //www.southampton.ac.uk/
-    image: /images/sponsors/university-of-southampton.svg
-
   - name: Adventurous Machines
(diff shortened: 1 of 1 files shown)


RealOrangeOne (approved):
Suggestion: Maybe worth putting a comment above `sponsors:` stating that it should be alphabetical?

sedders123:
Wasn't in the meeting where this was discussed, but is it possible for us to just decide the order so we only need to change once? I'm envisioning a question like "Why was I first, then second for a bit, then first again". Personally, I think ordering should be something we aim to change as infrequently as possible. 

In a previous PR I believe the order of when a sponsor started supporting us was used as the temporary ordering (https://github.com/srobo/website/pull/307#discussion_r715071254) until a decision has been properly made. Is there a reason alphabetical is better as a temporary solution?

antoinepetty:
Leaving this unchanged until an order is concretely decided

PR #312, opened 2021-10-05
Deprecate the kitbook

RealOrangeOne:
https://github.com/srobo/kitbook/commit/2bffe3f44405e31997d6b6679572d932a336c971

### DIFF
--- a/_env/nginx.conf
+++ b/_env/nginx.conf
@@ -156,7 +156,7 @@ http {
     # The runbook has replaced trac
     rewrite ^/trac                 /runbook/ permanent;
     
-    # The kitbook has been incorperated back into the runbook
+    # The kitbook has been incorporated back into the runbook
     rewrite ^/kitbook              /runbook/ permanent;
 
     # We have renamed the news section to blog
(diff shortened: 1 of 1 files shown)


trickeydan (approved):

sedders123:
On second thoughts, should this redirect to the runbook, so any links in the wild still work?

PeterJCLaw:
> On second thought, should this redirect to the runbook so any links in the wild still work?

That would presumably depend on the same sub-page existing in the runbook? A placeholder page which handles `/kitbook*` and tells people about the move might be an alternative?

sedders123:
I don't feel super strongly about it was just a thought, happy to go with consensus. Redirecting any kitbook route to the homepage of runbook feels like a low effort win. Given the relatively short lifespan of the kitbook I can't imagine there are too many links in the wild but the redirect would mean someone ends up at a semi useful page rather than a 404.

raccube (approved):

_env/nginx.conf
```diff
@@ -160,6 +155,9 @@ http {
 
     # The runbook has replaced trac
     rewrite ^/trac                 /runbook/ permanent;
+    
+    # The kitbook has been incorperated back into the runbook
```
sedders123, line 159:
```suggestion
    # The kitbook has been incorporated back into the runbook
```
```suggestion
    # The kitbook has been incorperated back into the runbook
```

PR #313, opened 2021-10-13
Add SR2022 kickstart event

RealOrangeOne:
Not a huge amount of detail yet, that'll come shortly. Just getting a link ready.

### DIFF
--- a/_events/sr2022/kickstart.md
+++ b/_events/sr2022/kickstart.md
@@ -1,0 +1,20 @@
+---
+title: Kickstart
+date: 2021-10-13 13:00:00 Europe/London
+layout: event
+type: kickstart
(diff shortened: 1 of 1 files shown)


antoinepetty (approved):

_events/sr2022/kickstart.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Kickstart
+date: 2021-10-13 13:00:00 Europe/London
```
PeterJCLaw, line 3: This is today. Is that intentional? It means that this event isn't showing up on the website.

RealOrangeOne: Ah damn it. No, no it is not.

Issue #290, opened 2021-09-01
Urls without trailing slashes redirect via insecure HTTP


PeterJCLaw:
For example the homepage at https://studentrobotics.org contains a link to `/volunteer` which, as you'd expect, goes to https://studentrobotics.org/volunteer. However _that_ page is then served as a redirect to http://studentrobotics.org/volunteer/ (adding the trailing slash, but removing the TLS).

In my browser config (Firefox, with HTTPS-Only Mode enabled) this means I get a warning when clicking the link.

I had thought we'd seen and fixed this in the past, but I can't recall the exact details nor the fix.

```
$ curl -I https://studentrobotics.org/sponsor
HTTP/1.1 301 Moved Permanently
server: nginx/1.21.1
location: http://studentrobotics.org/sponsor/
```

PeterJCLaw:
Think I found where I was thinking we'd seen this before: https://github.com/srobo/website/issues/122#issuecomment-457387477

PeterJCLaw:
Not sure what changed (since the links haven't) but this has stopped showing issues for me in Firefox :thinking:

sedders123:
Think this has been resolved with the move to `monty`

```
$ curl -I https://studentrobotics.org/sponsor
HTTP/1.1 301 Moved Permanently
Server: nginx/1.18.0 (Ubuntu)
Date: Thu, 03 Mar 2022 20:34:37 GMT
Content-Type: text/html
Location: https://studentrobotics.org/sponsor/
Connection: keep-alive
permissions-policy: interest-cohort=()
x-origin-cache: HIT
Access-Control-Allow-Origin: *
expires: Thu, 03 Mar 2022 20:44:37 GMT
Cache-Control: max-age=600
x-proxy-cache: MISS
X-GitHub-Request-Id: 2E30:E6FB:266CEF0:27BB0D7:6221265D
Accept-Ranges: bytes
Via: 1.1 varnish
Age: 0
X-Served-By: cache-lon4269-LON
X-Cache: MISS
X-Cache-Hits: 0
X-Timer: S1646339677.448991,VS0,VE102
Vary: Accept-Encoding
X-Fastly-Request-ID: e8885eaae1d9881ffde305832b6572af5abc226e
Strict-Transport-Security: max-age=15768000; includeSubDomains; preload
X-Frame-Options: SAMEORIGIN
```

PR #318, opened 2021-10-14
Improve links and avoid plain-http redirects

PeterJCLaw:
This changes links whose target isn't in this repo to being fully absolute, as if they were a separate website. This reflects the fact that these kinda are on separate websites and will make things easier for a link checker tool (see https://github.com/srobo/tasks/issues/701) to validate.

It also fixes some internal links where trailing slashes were missing and thus hitting plain HTTP redirects (see #290 which this helps but does not fix).

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -4,8 +4,8 @@
       <h4>For Competitors</h4>
       <ul class="sitemap">
         <li><a href="{{ '/teams/' | prepend: site.baseurl }}">Teams Dashboard</a></li>
-        <li><a href="/ide/">IDE</a></li>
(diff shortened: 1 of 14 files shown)


antoinepetty (approved):
I assume you'd need a matching footer for the docs and competition website?

PeterJCLaw:
Yeah, if this is a pattern we're happy with we should extend it to the other sub-sites.

PeterJCLaw:
For https://github.com/srobo/docs/pull/215 I've exploring using `{{ site.url }}` rather than our literal domain name, which feels like it might be slightly nicer here.

Issue #345, opened 2021-12-02
Remove teams dashboard, replace with docs button


antoinepetty:
This was discussed in a marketing team meeting. 

The only thing the page really adds to the site is clear links to the docs, when it had the IDE and forum links it made sense but now its just docs we could achieve the same goal with a top level nav item.

antoinepetty:
Also makes #340 obselete

PR #354, opened 2021-12-10
Show latest blog post and next event above the fold

sedders123:
At the moment these are only accessible by scrolling a long way down the page or through the Teams dashboard (which #345 removes).

With an upcoming event
image

Without an upcoming event
image

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -17,5 +17,30 @@ layout: default
     </div>
   </div>
 </div>
+<div class="panel panel-thin latest-content">
+  <div class="fixed-width">
+    <div class="content">
+      {% assign latest_post = site.posts.first %}
+      <h4>Latest Post:</h4>
+      <a href="{{ latest_post.url | prepend: site.baseurl }}">{{latest_post.title }}</a>
+      <div class="content-details">
+        {{ latest_post.date | date: "%e %B, %Y" }}
+      </div>
+    </div>
+    <div class="content">
+      {% assign next_event = site.events | where_exp: "event", "event.date > site.time"
+      | where_exp: "event", "event.cancelled != true" | first %}
+      <h4>Next Event:</h4>
+      {% if next_event %}
+        <a href="{{ next_event.url | prepend: site.baseurl }}">{{ next_event.title }}</a>
+        <div class="content-details">
+          {{ next_event.date | date: "%e %B, %Y" }}
+        </div>
+      {% else %}
+        <p>No upcoming events</p>
+      {% endif %}
+    </div>
+  </div>
+</div>
 {{ content }}
 {% include footer.html %}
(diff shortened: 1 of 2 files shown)


_layouts/home.html
```diff
@@ -17,5 +17,30 @@ <h2>Inspiring the next generation of engineers</h2>
...
+      {% assign next_event = site.events | where_exp: "event", "event.date > site.time"
+      | where_exp: "event", "event.cancelled != true" | first %}
+      <h4>Next Event:</h4>
+      {% if next_event %}
+        <a href="{{ next_event.url | prepend: site.baseurl }}">{{ next_event.title }}</a>
+        <div class="content-details">
+          {{ next_event.date | date: "%e %B, %Y" }}
+        </div>
+      {% else %}
+        <p>No upcoming events</p>
```
PeterJCLaw, line 40: Thoughts on whether this should include a link to the events page to see past events?

_sass/panel.scss
```diff
@@ -90,3 +90,32 @@ img.image {
 .panel-fill {
   width: 100%;
 }
+
+.panel.latest-content {
+  padding: 1rem 0;
+
+  .fixed-width {
+    display: flex;
+    justify-content: space-around;
```
PeterJCLaw, line 99: If the title of the blog post gets really long I think this will result in that taking up a lot more width than the event.
Is this actually the visual we want?
(My assumption is that we're aiming for them to each be centred in their own column and occupy half the width)
Maybe worth trying with one of the longer post-competition blog posts as the latest and seeing how it looks.

PeterJCLaw (commented):
:+1: do doing this, some suggestions about the implementation though.

_layouts/home.html
sedders123: 👍 e3f630b

_sass/panel.scss
sedders123: Yeah with longer blog posts titles the width split was indeed not 50/50. f33ca38 addresses this

_layouts/home.html
```diff
@@ -17,5 +17,32 @@ <h2>Inspiring the next generation of engineers</h2>
     </div>
   </div>
 </div>
+<div class="panel panel-thin latest-content">
+  <div class="fixed-width">
+    <div class="content">
+      {% assign latest_post = site.posts.first %}
+      <h4>Latest Post:</h4>
```
PeterJCLaw, line 24: minor: thoughts on not including the colon here? These are headings rather than phrases, so I think it might look better without the colons.

PeterJCLaw (dismissed):

_layouts/home.html
sedders123: 👍 2268b7c

PeterJCLaw (approved):

PR #386, opened 2022-03-18
Update VSCode Extension suggestions

PeterJCLaw:
This includes the spellchecker which is compatible with the one used in CI and a markdown linter I've found useful locally.
The other two are just generally useful, though less specifically applicable here.

Also removes the settings for spell checking extension no longer recommended.


sedders123 (approved):

raccube:
Thought: Is there an extension that shows a list of Rake tasks in the editor?

PeterJCLaw:
> Thought: Is there an extension that shows a list of Rake tasks in the editor?

Maybe? Feel free to find one and open a PR to add it.

PR #393, opened 2022-03-26
Update homepage statistics

sedders123:
These values more closely reflect the current state of SR


Scarzy:
Those do indeed look like numbers, and are plausible in line with what has happened over the past couple of years

Scarzy (approved):
Looks ok

PR #396, opened 2022-03-31
Update COVID guidance before the upcoming techday

Scarzy:
In the absence of information from the venues, and recognising that we need something in place by the tech day this weekend, I suggest we add this COVID policy. I am expecting it to change to some degree before the competition.

### DIFF
--- a/covid-19.md
+++ b/covid-19.md
@@ -8,9 +8,10 @@ We will keep this page updated with new information as it becomes available.
 
 Last updated: {{ page.date | date: "%e %B %Y" }}.
 
-1. Anyone who is experiencing symptoms of COVID-19, or has tested positive for COVID-19 in the 7 days beforehand (or lives with someone who has) must not attend.
+1. If you are experiencing symptoms of COVID-19, or have tested positive for COVID-19 in the 7 days beforehand (or live with someone who has) please do not attend.
 2. If you are able, please wear a mask when not in your team pit.
 3. Please clean your hands regularly, using hand sanitizer or soap.
+4. Be respectful towards others personal space.
 
 ## The Competition - SR2022
 
--- a/covid-19.md
+++ b/covid-19.md
@@ -11,7 +11,7 @@ Last updated: {{ page.date | date: "%e %B %Y" }}.
 1. If you are experiencing symptoms of COVID-19, or have tested positive for COVID-19 in the 7 days beforehand (or live with someone who has) please do not attend.
 2. If you are able, please wear a mask when not in your team pit.
 3. Please clean your hands regularly, using hand sanitizer or soap.
-4. Be respectful towards others personal space.
+4. Be respectful towards others' personal space.
 
 ## The Competition - SR2022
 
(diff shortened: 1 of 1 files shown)


covid-19.md
```diff
@@ -1,13 +1,18 @@
...
 ---
 
 We will keep this page updated with new information as it becomes available.
 
 Last updated: {{ page.date | date: "%e %B %Y" }}.
 
+1. If you are experiencing symptoms of COVID-19, or have tested positive for COVID-19 in the 7 days beforehand (or live with someone who has) please do not attend.
+2. If you are able, please wear a mask when not in your team pit.
+3. Please clean your hands regularly, using hand sanitizer or soap.
+4. Be respectful towards others personal space.
```
PeterJCLaw, line 14: nit:
```suggestion
4. Be respectful towards others' personal space.
```

covid-19.md
```diff
@@ -1,13 +1,18 @@
 ---
 layout: blog
 title: COVID-19 Updates
-date: 2021-12-08
+date: 2022-03-31
 ---
 
 We will keep this page updated with new information as it becomes available.
 
 Last updated: {{ page.date | date: "%e %B %Y" }}.
```
PeterJCLaw, line 10: Should we have a title or something here, so that it doesn't just launch straight into the list?
Even just `## Events` might do?

Scarzy: I'll add some prose

covid-19.md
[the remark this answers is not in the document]
Scarzy: Do you have a screenshot?

covid-19.md
Scarzy: I changed most of the wording really...

covid-19.md
```diff
@@ -1,15 +1,26 @@
...
+## SR2022
 
-It is currently our intention to run an in person competition for the SR2022 competition cycle. This is of course dependent on the government guidelines which we are keeping a close eye on. If you have any COVID-19 specific questions about the competition please contact us at [{{ site.emails.teams }}][teams-email].
+Life is beginning to get back to normal, and this means we are finally able to do in person events again!
+However, we must recognise that some people still have to be cautious, to protect their own health or their friends or family.
+As such, we ask that everyone follows the below requests to help ensure we all keep safe, whilst still having lots of fun with robots!
+
+* If you are experiencing symptoms of COVID-19, or have tested positive for COVID-19 in the 7 days beforehand (or live with someone who has) please do not attend.
+* If you are able, please wear a mask when not in your team pit.
+* Please clean your hands regularly, using hand sanitizer or soap.
```
RealOrangeOne, line 19: Worth mentioning that we won't be providing sanitizer?

Scarzy: I thought about this.

I'm largely still debating whether we need to provide some or not (at least for the main competition). However that's beyond the scope of this PR.

I can't say with any confidence whether or not the venue for either the techday or the competition will have hand sanitizer available anyway.
It is reasonable to assume the toilets will have some kind of soap or sanitizer in them, which means there is always the soap fallback.
Some young people can, at times, be incredibly pedantic and might use any kind of inaccuracy to justify not complying.
A number of people will carry their own anyway.
So it's easier to just not mention it.

PR #400, opened 2022-04-10
Add the concept of an end date for events

RealOrangeOne:
image

Not entirely happy with the layout, but it's very simple.

~~I wanted to be able to just show the end time if they're on the same date, but I couldn't get the templating to work. If someone else can (such as future me), then please do tell me how to do it!~~

I've tested that the resulting calendar event looks correct: 
image


events.ics
```diff
@@ -8,7 +8,8 @@ METHOD:PUBLISH
 {% for event in sorted %}BEGIN:VEVENT
 UID:{{ event.url | prepend: site.baseurl | prepend: site.url }}
 DTSTART;TZID={{ site.timezone }}{% if event.time_tbc %};VALUE=DATE:{{ event.date | date: "%Y%m%d" }}{% else %}:{{ event.date | date: "%Y%m%dT%H%M%S" }}{% endif %}
-DTSTAMP{% if event.time_tbc %};VALUE=DATE:{{ event.date | date: "%Y%m%d" }}{% else %}:{{ event.date | date: "%Y%m%dT%H%M%S" }}{% endif %}
+{% if event.end_date %}DTEND;TZID={{ site.timezone }}{% if event.time_tbc %};VALUE=DATE:{{ event.end_date | date: "%Y%m%d" }}{% else %}:{{ event.end_date | date: "%Y%m%dT%H%M%S" }}{% endif %}{% endif %}
+DTSTAMP:{{ event.date | date: "%Y%m%dT%H%M%S" }}
```
PeterJCLaw, line 12: This appears to lose some of the support for `time_tbc` -- could we validate that that still works?

_includes/datetime.html
```diff
@@ -9,6 +9,17 @@
   >
     {{ include.event.date | date:include.date_only_format }}
   </time>
+  {% if include.event.end_date %}
+    &mdash;
+    <time
+      {{ cancelled }}
+      class="time-span"
```
PeterJCLaw, line 16: Thoughts on making `time-span` prevent wrapping of the contained text? I Suspect that might look slightly better than the current wrapping result.

events.ics
RealOrangeOne: `DTSTAMP` is the creation time for the event, so it doesn't really need the concept of `time_tbc`. ea3c018a179b3c17ba4956cd91ec1ae6949270a9

PeterJCLaw:
> I wanted to be able to just show the end time if they're on the same date, but I couldn't get the templating to work. If someone else can (such as future me), then please do tell me how to do it!

https://github.com/RealOrangeOne/srobo-website/pull/1 has a partial approach at this. Still WIP as it doesn't handle localisation. I hope to get to that later today.

_includes/datetime.html
PeterJCLaw, line 16: I don't think this should be a `time-span`. That class is used by the localisation JS to localise the times of the event. If we're in the `time_tbc` branch there isn't a time set yet and so there's nothing to localise.
```suggestion
```

RealOrangeOne: image

I agree this looks better, but I can't decide whether the trailing dash looks nice? Thoughts?

PeterJCLaw: I think that's preferable to the current wrapping.
IMO extra good would be that the end datetime is indented to align with the start (though that would look super weird if it was just the time that wrapped), so probably not worth it.

PeterJCLaw (approved):

PR #403, opened 2022-04-15
Link from the event page to the competition matches pages

PeterJCLaw:
Probably best not to merge this until https://github.com/srobo/sr2022-comp/pull/6 is merged as currently the knockouts show teams placed in the first round of matches which likely aren't correct.


RealOrangeOne (approved):
Assuming these links still work when in the "competition homepage" mode

PeterJCLaw:
> Assuming these links still work when in the "competition homepage" mode

Yes. The only thing which changes in that mode is the content served at the root url.

PR #405, opened 2022-04-15
Prevent any time elements from wrapping

RealOrangeOne:
We can't just target the ones on the event page, as they also exist on the event listing page.

In reality, we probably don't want them ever to wrap.

### DIFF
--- a/_sass/common.scss
+++ b/_sass/common.scss
@@ -63,3 +63,7 @@
   display: flex;
   justify-content: center;
 }
+
+time {
+  white-space: nowrap;
+}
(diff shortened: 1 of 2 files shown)


sedders123 (approved):

PeterJCLaw (commented):
I'd suggest not using this approach -- it could result in weirdness if a `<time>` element was used as part of prose.

Instead I suggest a class applied to the time elements by the `datetime.html` include.
We could even introduce a generic `text-nowrap` helper and use that.

PeterJCLaw:
@RealOrangeOne suggest we close this unless you're planning to pick it up soon?

PeterJCLaw:
Closing this for now. Feel free to re-open if/when this is needed.

PR #407, opened 2022-04-15
Ensure variables are cleared if they're not needed

RealOrangeOne:
Turns out these assignments are not scoped to the template they are defined in, and get reused with future calls when rendering the same page

The effect of this issue can be seen on the events page, where it shows non-cancelled events with the cancelled styling

image


PeterJCLaw (approved):
Good spot!

PR #426, opened 2022-08-04
Rename sponsor-logo class to avoid adblock

PeterJCLaw:
It feels reasonable that we do this as these aren't advertisers -- they are genuinely sponsors of us as a charity, so aren't the kind of thing that adblock extensions would normally aim to hide.

Additionally these content are served entirely from our own hosting, so there's no user-tracking implications from these being shown.

If we're happy with this approach I'll copy it over to https://github.com/srobo/competition-website/ too

### DIFF
--- a/_sass/sponsors.scss
+++ b/_sass/sponsors.scss
@@ -109,7 +109,8 @@ ul.sponsor-list {
   }
 }
 
-.sponsor-logo {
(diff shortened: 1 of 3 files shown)


antoinepetty:
Thanks Peter, this works for me with adblock on Chrome :)

antoinepetty (approved):

PR #441, opened 2022-10-02
Change references of team leaders to team supervisors

sedders123:


PeterJCLaw (approved):
Do you want to also update the old blog posts & events for consistency?

sedders123:
I left old references (there are only three I found, all events) as they use the correct terminology for their time and I'm not sure if it really would save enough confusion to warrant changing them (don't think many would scroll back through to events from 2020 or 2017).

PR #460, opened 2022-11-15
No power tools at all

PeterJCLaw:
Hopefully fixes a mistaken implication from #459.

### DIFF
--- a/_events/sr2023/london-tech-day-december.md
+++ b/_events/sr2023/london-tech-day-december.md
@@ -22,7 +22,7 @@ like to attend. We will be confirming places shortly after you sign up.
 At the very latest please signup by Wednesday 30th November 2022, though we may
 close signups sooner if all places are filled.
 
-At the request of the venue: paint, power tools or hand tools which create mess
+At the request of the venue: paint, hand tools which create mess or power tools
 may not be used at this Tech Day.
 
 **Please bring your laptops!**


_events/sr2023/london-tech-day-december.md
```diff
@@ -22,7 +22,7 @@ like to attend. We will be confirming places shortly after you sign up.
 At the very latest please signup by Wednesday 30th November 2022, though we may
 close signups sooner if all places are filled.
 
-At the request of the venue: paint, power tools or hand tools which create mess
+At the request of the venue: paint, hand tools which create mess or power tools
```
sedders123, line 25: Nit: Would this clear up the confusion about the "hand tools which create power tools"?
```suggestion
At the request of the venue: paint, hand tools which create mess, or power tools
```

Also happy for this to merged as is if you'd rather

sedders123 (approved):

_events/sr2023/london-tech-day-december.md
PeterJCLaw: Ah, hrm. :thinking: I agree that solves the "hand tools which create power tools" but the grammar pedant inside me doesn't like it.

I think I'm going to ship without this, but paging @prophile in case he can come up with something better.

sedders123: Grammarly suggests its ok https://www.grammarly.com/blog/comma-before-or/

Would an alternative be to change the `or` to `and`

PeterJCLaw: 1380f07c9e1917e594cf95971ec5aa72d6ea9fc2

PR #486, opened 2023-02-08
Add an "add to calendar" button to event pages

RealOrangeOne:
This adds an "Add to calendar" button to the events list and event page. The buttons always try to embed the full event list, rather than just adding the single event. It's a little misleading, but it ensures the person sees all our future events too.

image

### DIFF
--- a/_layouts/event.html
+++ b/_layouts/event.html
@@ -1,9 +1,10 @@
 ---
 layout: page
 ---
-<div class="header-card">
+<div class="header-card calendar-button">
   <div class="fixed-width">
     <h1>{{ page.title }}</h1>
+    {% include calendar-button.html %}
   </div>
 </div>
 <div class="main">
--- a/_includes/calendar-button.html
+++ b/_includes/calendar-button.html
@@ -1,0 +1,12 @@
+<add-to-calendar-button
+  name="{{ site.title }}"
+  {% comment %}startDate is required{% endcomment %}
+  startDate="1970-01-01"
+  subscribe
+  icsFile="https://studentrobotics.org/events.ics"
+  iCalFileName="{{ site.title }}"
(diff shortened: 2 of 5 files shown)


_layouts/event.html
```diff
@@ -1,9 +1,10 @@
 ---
 layout: page
 ---
-<div class="header-card">
+<div class="header-card calendar-button">
```
sedders123, line 4: This `div` is not a `calendar-button`. Naming this one seems a little tricky, `calendar-button-container` maybe? Though looking at they styling a `header-card-with-calendar-button` may be better? Not 100% sure so open to alternatives

sedders123:
Can we make it not blur the entire page when hovered over? Not sure why we'd to obscure the information about the event just because a user is potentially about to add it to their calendar?

Also it seems to respect the system theme which while cool means I have a black button in the blue header when we should probably just force it to be white
image

sedders123 (changes requested):
When the dropdown is open the mouse cursor is changed to an SVG, can we keep the default cursor please. Also can we reduce the border on `.atcb-list` to be `1px` rather than the `0.15em` it is now.

Also not a fan of the checkmark that is displayed once the user clicks an option within the dropdown
image
could we please remove this?

Do we actually need a third party dependency for this functionality? A dropdown with links to add our calendar to those providers could be much more lightweight for us to implement ourselves. We don't seem to want the majority of the features this component provides. (Also, personally, I'm mildly irritated at how this component uses JS to open links rather than `<a>` tags. Means you can't middle click to open in new tab, yes left clicking does open in a new tab but it breaks my assumptions as a user of the web)

sedders123:
On mobile I think it'd be nice if this displayed inline with the content rather than causing the header to squish?

<details>
<summary>Current</summary>

image
</details>

<details>
<summary>Proposed (Excuse slightly jank mockup)</summary>

image
</details>

PR #493, opened 2023-03-04
Fix entrance location on Boldrewood pdf

Adimote:


sedders123:
The new map points at building 177. Is that the correct the description says 176?
image

Adimote:
> The new map points at building 177. Is that the correct the description says 176? image

nope, will fix in this PR too

Adimote:
Now fixed @sedders123  :)

sedders123 (approved):

PR #494, opened 2023-03-13
SR2023 competition risk assessment

Scarzy:
Add the SR2023 competition risk assessment

(Contents of the RA reviewed separately, please only review the adding of the document to the website)


resources/sr2023/risk-assessments/competition_ra_southampton.pdf
PeterJCLaw, line 1: Would be great to include some details of the origin of this file in the commit (or at least the PR description). For files which came from a repo build or the like, we'll typically reference the commit and CI job which built the binary.
This aids with understanding the history of binary blobs as well as easing review, since the reviewer can trivially validate that the file is indeed the build output from the upstream.

PeterJCLaw, line 1: Not sure if intentional -- the "COVID-19" row is on a page on its own. Maybe tweak the margins a bit so it can flow onto the previous page?

PeterJCLaw (approved):
:+1:

resources/sr2023/risk-assessments/competition_ra_southampton.pdf
Scarzy: That's just Google formatting it weirdly. As a couple of schools have asked for it, I'll worry about fixing that at a later date (risk assessments should be able to adapt over time anyway)

Scarzy: Good point
PDF export of the document in the competition google drive

PeterJCLaw: Can we link to said doc?

PR #501, opened 2023-04-12
Comment out Competition Status part of home page

sedders123:
Not to be merged before 26th (want to let people see scores for a bit before we remove)


prophile (approved):

PeterJCLaw:
I suggest that we use a configuration setting for this rather than commenting it out. That would more easily allow us to play with it during the year and clarifies that it's a part of the site which should be maintained rather than allowing it to degrade. Alternatively I would encourage that we just remove this. Commented out code tends to drift from being useful, so is usually best avoided.

PR #506, opened 2023-05-11
Add subresource integrity checking to our user-tracking script

PeterJCLaw:
If we're going to track users we should at least ensure that the script which does so is what we're expecting. The value here is derived from a manual check of the current version of the file.

I couldn't find anything on Plausible's website about this, so I'm just guessing that this is ok to do. Ideally it feels like we should pin the version of the library we're pulling in, however I couldn't find anything on how to do that either.

### DIFF
--- a/_includes/script.html
+++ b/_includes/script.html
@@ -2,5 +2,5 @@
 <script src="{{ '/js/localise-times.js' | prepend: site.baseurl }}"></script>
 
 {% if site.user_tracking %}
-  <script defer data-domain="studentrobotics.org" src="https://plausible.io/js/plausible.js"></script>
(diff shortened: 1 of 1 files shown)


RealOrangeOne (changes requested):
This will silently break functionality if the script is changed upstream (which definitely happens as the Plausible application is deployed).

SRI isn't something officially supported, but there are discussions about it: https://github.com/plausible/analytics/discussions/380

PeterJCLaw:
Hrm, I did suspect that that failure mode might be the case. Really not a fan of their approach that it's too hard to actually version it, especially given that they do already offer an [NPM package](https://www.npmjs.com/package/plausible-tracker) which itself is already (and must be) versioned!

Perhaps we should pull in the NPM package instead? We could fetch that via https://www.jsdelivr.com/ perhaps or maybe even move towards building the site ourselves (at which point we have more freedom for other things anyway).

PR #511, opened 2023-06-18
Update UoS logo

RealOrangeOne:


PeterJCLaw:
@RealOrangeOne please could you edit the commit to include some details of the provenance of the logo? It _looks_ right to me, but a link to where you got it from would be great (both it would help us review that this change is correct and it'll make things easier in future as we'll know where to look for updated versions). Including both the actual url to the source and any links to docs from the Uni on the usage of their logo would be excellent.

sedders123 (approved):

sedders123:
For provenance I pointed Jake at an SVG I had downloaded from the Uni's website and Jake removed some of the oddities. Merging this now so that the logo is correct on the website

PR #515, opened 2023-06-28
Override updating configstore

PeterJCLaw:
There's a transitive dependency from `configstore` to a version of `semver` which has a [ReDOS issue](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw). While it's unlikely to be an issue for us this fixes a dependabot report, which makes things easier.

If we're happy with this I'll do the same to our other repos which use `cspell`.


PeterJCLaw:
This doesn't work due to ESM stuff.

PR #519, opened 2023-07-15
Add Robots 101 list page

RealOrangeOne:
For pointing supervisors how to get started, our "Robots 101" series is great, but there currently it's a way to link them to the entire series of posts.

This leans on Jekyll's tag functionality, and uses a custom list page to show the posts. To keep the page smaller, only the page description is shown.

image

Opening early for feedback, there's probably more polish which could be done.

### DIFF
--- a/_layouts/blog.html
+++ b/_layouts/blog.html
@@ -8,7 +8,9 @@ layout: page
         <div class="article">
           <div class="article-header">
             <h1>{{ page.title }}</h1>
-            <p class="article-info">Published {{ page.date | date: "%e %B, %Y" }}</p>
+            {% if page.date %}
+              <p class="article-info">Published {{ page.date | date: "%e %B, %Y" }}</p>
+            {% endif %}
           </div>
           <div class="article-content">
             {{ content }}
--- a/blog/robots-101.html
+++ b/blog/robots-101.html
@@ -1,0 +1,20 @@
+---
+layout: blog
+title: Robots 101
+---
+
+<p>Welcome to Robot 101, our series where we guide you through the competition and give some advice on your next steps.</p>
+
+{% for page in site.tags.robots_101 %}
+  {% include get_description_for_page.html %}
+
+  <div class="article article-preview card">
+    <div class="article-header">
+      <h1><a href="{{ page.url | prepend: site.baseurl }}">{{ page.title }}</a></h1>
+      <p class="article-info">Published {{ page.date | date: "%e %B, %Y" }}</p>
+    </div>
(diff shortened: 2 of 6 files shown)


_layouts/blog.html
```diff
@@ -8,7 +8,9 @@
         <div class="article">
           <div class="article-header">
             <h1>{{ page.title }}</h1>
-            <p class="article-info">Published {{ page.date | date: "%e %B, %Y" }}</p>
+            {% if page.date %}
+              <p class="article-info">Published {{ page.date | date: "%e %B, %Y" }}</p>
+            {% endif %}
```
PeterJCLaw, line 13: This change seems unrelated?
Also: when might a page not have a date?

blog/robots-101.html
```diff
@@ -0,0 +1,20 @@
...
+
+<p>Welcome to Robot 101, our series where we guide you through the competition and give some advice on your next steps.</p>
+
+{% for page in site.tags.robots_101 %}
+  {% include get_description_for_page.html %}
+
+  <div class="article article-preview card">
+    <div class="article-header">
+      <h1><a href="{{ page.url | prepend: site.baseurl }}">{{ page.title }}</a></h1>
+      <p class="article-info">Published {{ page.date | date: "%e %B, %Y" }}</p>
```
PeterJCLaw, line 14: Maybe omit the publication date from this list page? If the idea is that we can use these articles as evergreen resources for team supervisors, having the overview show them as being old feels undesirable.

(I don't think we can remove the dates entirely, but this is probably where they're most visible)

blog/robots-101.html
```diff
@@ -0,0 +1,20 @@
+---
+layout: blog
+title: Robots 101
+---
+
+<p>Welcome to Robot 101, our series where we guide you through the competition and give some advice on your next steps.</p>
+
+{% for page in site.tags.robots_101 %}
+  {% include get_description_for_page.html %}
```
PeterJCLaw, line 9: I'm not sure that using the _description_ here is quite right? This is intended to be part of the metadata for a page, exposed in header information. It falls back to the site description for example.

I suggest instead that we use `page.excerpt` directly, processing it as needed.

I also wonder if it would be worth us introducing a span or something around the "welcome to the series" blocks on each page which this can then skip over to avoid repetition and give more of a feel for the actual page content.

raccube:
I’d like to at some point get these into the docs.

PeterJCLaw:
Let's drop this now these posts are in the docs.

PR #524, opened 2023-09-25
Close signups for SR2024

sedders123:
This removes the signup form and swaps back to the default MailChimp contact details form.


Adimote (approved):

sedders123:
Content is the exact same but force pushed for a slightly better git history

PR #536, opened 2023-11-02
Build and publish in CI

raccube:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  push:
+    branches:
+      - main
+      - ci
+  schedule:
+    - cron: 0 0 * * *
```
PeterJCLaw, line 7: I suggest building each day slightly after midnight so that we don't have race conditions. We might also want to work out what happens during BST -- whether this ends up being the previous day.

.github/workflows/validate.yml
```diff
@@ -1,29 +0,0 @@
...
-    branches:
-      - main
-  pull_request: {}
-
-jobs:
-  validate:
-    runs-on: ubuntu-latest
-    env:
-      # ruby/setup-ruby@v1 does not offer a way to change the cached gems path.
-      # See https://github.com/ruby/setup-ruby/issues/291
```
PeterJCLaw, line 14: Please can we preserve this explanatory comment. (Even though the issue has been closed as won't fix, having the reference is useful to explain what's going on here)

.github/workflows/validate.yml
```diff
@@ -1,29 +0,0 @@
-name: Validate
-
-on:
-  push:
-    branches:
-      - main
-  pull_request: {}
```
PeterJCLaw, line 7: I think we should keep this? This is the part which tells GitHub Actions to run on each PR.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  push:
+    branches:
+      - main
+      - ci
```
PeterJCLaw, line 5:
```suggestion
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
...
+  schedule:
+    - cron: 0 0 * * *
+name: Build & Publish
```
PeterJCLaw, line 8: style: Please can we put the `name` at the top and have an empty line between each top-level entry in the YAML file.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          cache-version: 2
```
PeterJCLaw, line 23: I think this will need to specify the version explicitly to stay on the right version (2.7) for now?

raccube: No, it reads from `.ruby-version`

raccube:
We will need to change the required job to not be build and not validate

PeterJCLaw (approved):

PR #537, opened 2023-11-04
added docker implementation to the website

Muhie:
- changed readme to show that you now can use docker instead of only being able to use ruby


Muhie:
changed readme to show that you now can use docker instead of only being able to use ruby

PR #560, opened 2024-03-30
Add SR2024 playlists

RealOrangeOne:
Fixes https://github.com/srobo/tasks/issues/1270


PeterJCLaw (approved):
Confirmed these links go to something which looks right.
Do we need to check that e.g: unexpected people can't add things to the lists or anything?

RealOrangeOne:
Playlists are read-only by default, and requires an invite link to edit. These links aren't the editing kind, so we're good.

(Better documentation for these playlists is on my list for some point)

PR #562, opened 2024-04-15
Fix the link to the image for the page

PeterJCLaw:
This is used in our header tags but not currently on the page itself, so this would have been affecting social media previews of the article.


RealOrangeOne (approved):
Perhaps a way to ensure this is to use `page.image` in the below `include`? (I think that's possible)

PeterJCLaw:
> Perhaps a way to ensure this is to use `page.image` in the below `include`? (I think that's possible)

Yeah, thought about that. Not sure we want to couple the two given the size of the image we want to embed, plus there may be other cases which better validation would solve for. Also wanted this to be a minimal quick fix.

PR #567, opened 2024-04-29
Remove reference to closed competition

RealOrangeOne:
THis could be construde as the competition _starting_ in 2024.

In future, we should remove this earlier.


PeterJCLaw (approved):
I think there might be a better wording we can use here which wouldn't need removing, but :shipit: for now and we can add something back if needed.

PR #610, opened 2025-03-25
Add SR2025 Venue Maps to the Competition event page

PeterJCLaw:
These are extracted from the CI build of the floorplans repo:
- https://github.com/srobo/comp-floorplans/commit/e0c7ce2d09fc6828c7ff85a17391e87b745cdce1
- https://github.com/srobo/comp-floorplans/actions/runs/13957873494/artifacts/2784492185

Files 'map-with-teams.{pdf,png}' renamed to 'venue-map.{pdf,png}'.

Contributes to https://github.com/srobo/tasks/issues/1426.


WillB97:
It would be better to take artifacts from the release.

PeterJCLaw:
> It would be better to take artifacts from the release.

Huh, hadn't spotted that there were releases on that repo. The commit in question however is the commit which is tagged as the latest release, so is it not safe to assume that the artefacts are the same? (I can't actually see how the artefacts get from the CI build to the release, though I'd strongly hope that they're not being modified in that process).

PeterJCLaw:
> > It would be better to take artifacts from the release.
> 
> Huh, hadn't spotted that there were releases on that repo. The commit in question however is the commit which is tagged as the latest release, so is it not safe to assume that the artefacts are the same? (I can't actually see how the artefacts get from the CI build to the release, though I'd strongly hope that they're not being modified in that process).

Comparing the assets: it seems that the PNG is the same, however the PDF is different. Do we know what's changed in the PDF on the release? (I can't immediately see any differences, though would very much prefer to use an asset which has clear provenance)

PeterJCLaw:
For clarity: I'm not planning to update this PR to use the assets from the release -- the files there appear to be in error vs CI and we should prefer CI generated assets.

WillB97 (changes requested):
Please use the artifacts from the release, I have just fixed the text rendering in the PDF.

Note: the artifacts of the release were generated from the action on the commit pointed to by the release, not the tag that was created by the release.

WillB97:
This action was used: https://github.com/srobo/comp-floorplans/actions/runs/13955524761.

I have manually fixed font rendering on the PDF using the SVG from this output.

PeterJCLaw:
What needed fixing? Both actions runs produce identical SVGs. All three PDFs also look fine to me.

If there's something wrong with how the CI produces the assets please could we at least record that as an issue on the repo along with the steps being used to address it? We shouldn't be relying on opaque manual steps as they make it harder for other people to understand processes and are likely to be forgotten in future. If we need to rely on them we should at least document the need and what they are.

PeterJCLaw:
Meta: would it be worth moving to a quick chat for this? I'm conscious that back & forths on a PR easily end up feeling antagonistic through trivial misunderstandings. I accept there are probably good reasons for your request here, I'm just trying to understand what it is and how we ensure things are clear for the future.

WillB97:
Capital A's render incorrectly.
image

PeterJCLaw:
That is most odd, though now you mention it I feel we've maybe had this issue before? Oddly it doesn't reproduce for me, even now I know what to look for, very weird. We should definitely get this logged so we're not relying on individuals happening to know about it to fix it.

PeterJCLaw:
https://github.com/srobo/comp-floorplans/issues/41

PeterJCLaw:
The competition happened.

PR #612, opened 2025-04-02
Link to the SR2025 venue map from the image too

PeterJCLaw:
This matches what we've done for this section in the past (the new source is copy/pasted with the year updated).

Also move the PDF link down to the footer as usual.

### DIFF
--- a/_events/sr2025/competition.md
+++ b/_events/sr2025/competition.md
@@ -71,9 +71,9 @@ We have a limited amount of free parking on site for competitors.
 
 ### Layout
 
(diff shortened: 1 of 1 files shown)


WillB97:
We don't do this with the paking image directly above.
Also, why are we linking it? People can just right-click and save-as on the image.

PeterJCLaw:
People _can_ right click & save it, but if they just want to see the file bigger then that doesn't necessarily help.
IMO we _should_ link to other images if they're meaningfully higher resolution than the page displays them (the parking image is barely bigger).
We have done this for the past few years at least, there doesn't seem to me to be a reason to stop doing so.

Alexbruvv (approved):

